PR #54, opened 2018-01-19
Update charts version to alpha12

alexberazouski:


vaadin-bom/pom.xml
```diff
@@ -18,7 +18,7 @@
     <flow.version>1.0.0.alpha18</flow.version>
     <testbench.version>6.0.0.alpha4</testbench.version>
     <board.version>1.1.0</board.version>
```
alexberazouski: board version seems to be obsolete

Artur-: yes, I will remove it or change it before beta

PR #58, opened 2018-01-22
Use flow lumo integration instead of just lumo themes.

SomeoneToIgnore:


SomeoneToIgnore:
> $ rake
rake aborted!
No Rakefile found (looking for: rakefile, Rakefile, rakefile.rb, Rakefile.rb)

PR #138, opened 2018-04-18
Use new vaadin spring artifacts

denis-anisimov:
Use Vaadin 10 Spring add-on.

### DIFF
--- a/vaadin-bom/pom.xml
+++ b/vaadin-bom/pom.xml
@@ -14,6 +14,7 @@
 
     <properties>
         <flow.version>1.0.0.beta7</flow.version>
+        <flow.spring.version>10.0.0.beta1</flow.spring.version>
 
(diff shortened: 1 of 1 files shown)


Artur-:
How is this artifact versioned? 10.0.0.beta1 is not in sync with anything?

denis-anisimov:
https://github.com/vaadin/flow/issues/3695

I have no idea.
The ticket says `10.0.0.beta1`. I made it `10.0.0.beta1`.
I'm not a decision maker.
@pleku ?

pleku:
@Artur- It was decided that we make a clear cap between the FW8 compatible spring add-on versions (3.0) and the V10 version. Thus we jumped to 10, which means that the first version of this artifact will be 10.0.0.beta1. Soon to be 10.0.0.rc1.

PR #169, opened 2018-05-17
Use testbench BOM and update to beta9

alvarezguille:
`vaadin-testbench-bom` is a new artifact that includes `vaadin-testbench-core` and the required selenium dependencies with correct versions

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/169)


alvarezguille:
We'll need to wait for beta10

PR #177, opened 2018-05-22
Remove Vaadin-spring-boot-starter from platform

ZheSun88:
According to the issue https://github.com/vaadin/spring/issues/298 , `vaadin-spring-boot-starter` is using platform dependency. so it should not be included in the `vaadin-bom`

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/177)


ZheSun88:
combine this with the flow update PR.

PR #204, opened 2018-06-08
Update template-bom

qtdzz:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/204)


qtdzz:
we can use --platform=10.0-SNAPSHOT, then `mvn versions:set` to the version we want. This PR is not needed then

PR #222, opened 2018-07-06
Update component versions and remove add-ons repo

### LINKED ISSUE #216: vaadin-bom references artifacts not available in Maven central
It would be [a good habit](https://maven.apache.org/repository/guide-central-repository-upload.html) for all artifacts published in maven central to reference only artifacts that are available in maven central. 

This causes issue like these:
https://github.com/spring-io/initializr/pull/694

alvarezguille:
Fixes #216  

This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/222)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,10 +38,6 @@
 
     <repositories>
         <repository>
-            <id>vaadin-addons</id>
-            <url>http://maven.vaadin.com/vaadin-addons</url>
-        </repository>
-        <repository>
             <id>vaadin-prerelease</id>
             <url>http://maven.vaadin.com/vaadin-prereleases</url>
         </repository>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -37,10 +37,6 @@
     </scm>
 
     <repositories>
-        <repository>
-            <id>vaadin-addons</id>
-            <url>http://maven.vaadin.com/vaadin-addons</url>
-        </repository>
         <repository>
             <id>vaadin-prerelease</id>
```
SomeoneToIgnore: Do we beed that one still?

mstahv: Yep, I think vaadin-prerelease repository shouldn't be there either.

alvarezguille: We don't need it now, but we will need to add it again when we start with V11 alphas, done.

PR #228, opened 2018-07-13
get rid of obsolete slf4j dependencies

mstahv:
We shouldn't need to declare external dependencies in our bom. This change is theoretically backwards incompatible, if somebody has depended on the slf4j version specified by our bom, so we should mention this in release notes.

Also we need to change bewerage buddy example, as that does exactly this.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/228)


Legioth:
Already handled in a different way through https://github.com/vaadin/platform/issues/243.

PR #262, opened 2018-08-06
Determine platform version and WebJars' dependencies automatically.

SomeoneToIgnore:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/262)


alvarezguille:
*[a discussion](https://reviewable.io/reviews/vaadin/platform/262#-LJwjyhd0Bls5dY1YWi0:-LJwjyhd0Bls5dY1YWi1:b-xq4y3q) (no related file):*
This PR is doing two changes related to bom generation
* Remove the need for platform argument
* Compute the transitive dependencies for bom
  * I'm a bit concerned about this para as there's no transparency about things being added or removed from bom, major version changes might be accidentally included in releases. It's quite hard to compare the generated result with generated in last release, until now it's possible to check `template-vaadin-bom`. In most cases this won't happen, but when it happen we probably won't notice

---

qtdzz:
An alternative solution is in https://github.com/vaadin/platform/pull/369

Issue #275, opened 2018-08-09
Generate javadoc for the vaadin artifact


ripla:
Generate and deploy javadoc for the vaadin artifact in addition to vaadin-core. The whole platform javadoc should be made available in vaadin.com/api. Related to #153 and #105.

alvarezguille:
what about testbench? 
vaadin artifact will not include javadoc for testbench (vaadin-testbench-core and vaadin-components-testbench) as it's not a dependency

ripla:
@alvarezguille good point, TB should be included. The end goal is to get the whole platform api available in one package so it can be consumed by vaadin.com/api. I think the actual mechanism on how to get that doesn't matter.

Legioth:
I guess we could define an internal meta module that depends on `vaadin` and `testbench` and generate javadocs based on that, assumed that we find a good way of getting javadocs for transitive dependencies.

PR #285, opened 2018-08-21
Generate javadoc for platform

qtdzz:
Fixes https://github.com/vaadin/platform/issues/275
Probably we should have a better module name, I can't come up with any better one.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/285)


alvarezguille:
we should carefully test with the release build staging these changes, at least in the past there was some issue with dependency fetching for javadoc when deploying to sonatype

PR #286, opened 2018-08-28
Update all components for V12, add Material

web-padawan:
Picked the proper versions of both free and commercial components.

Downgrades are undesirable as all the versions depend on same versions of
- `vaadin-development-mode-detector` 2.0.0
- `vaadin-usage-statistics` 2.0.0

The goal for the Components team is now to make beta versions by the end of September, and go to stable releases by the mid of October in order to make all this stuff for V12.

Note: another task would be to add P3 conversion step to platform publishing (to be done separately)

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/286)

PR #287, opened 2018-08-29
Update Flow, Components and Java versions

pleku:
Update to the latest prerelease versions for Flow, components and component Java integrations.

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -1,38 +1,39 @@
 {
     "core": {
         "flow": {
-            "javaVersion": "1.0.3"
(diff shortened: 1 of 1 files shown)


web-padawan:
Note: JS versions of components are already stable, and submitted the update in #286
We also started alpha period for components with renderers, and need to get in into V12

pleku:
I made the PR with matching versions of the components to the ones the Java wrappers have been built with. The wrappers will be updated to stable releases later on.

PR #352, opened 2018-10-18
Update vaadin-overlay to latest patch release

web-padawan:
Delivering a fix for using `vaadin-combo-box` in nested `vaadin-dialog`.


web-padawan:
@qtdzz PTAL

web-padawan:
@alvarezguille @qtdzz FYI, this change is especially important for the dialog component.
We should make sure the version is bumped for the next V12 alpha.

PR #401, opened 2018-11-13
Use version directly for license-checker

qtdzz:
Same as https://github.com/vaadin/platform/pull/400

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/401)


qtdzz:
Fixed in https://github.com/vaadin/platform/pull/409

PR #428, opened 2018-11-29
List components within a repository in release notes

ripla:
Some component repositories contain multiple components. Those components are not currently listed in the release notes. This PR adds those hidden components as a list in the release notes.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/428)


ripla:
@pleku thanks, I'll add those. I think the naming is a difficult question. We're referring to the "generic" component, so both `vaadin-button` and `Button` are wrong in their own way. I don't have any issues with the current naming, but I'll check around.

ripla:
Suggestion from @Artur-: list components, not repositories. I'll try that approach and gather feedback.

PR #493, opened 2019-01-10
Update v12 components versions

web-padawan:
Polymer update is needed for npm especially, as otherwise when installing `@vaadin/vaadin` user will get duplicated dependencies. I recommend stop pinning Polymer for npm completely.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/493)

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -22,19 +22,19 @@
         "vaadin-button": {
             "npmName": "@vaadin/vaadin-button",
             "javaVersion": "1.2.1",
-            "jsVersion": "2.1.1",
+            "jsVersion": "2.1.2",
             "component": true
         },
         "vaadin-checkbox": {
             "npmName": "@vaadin/vaadin-checkbox",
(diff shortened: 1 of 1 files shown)


versions.json
```diff
@@ -11,8 +11,8 @@
         },
         "polymer": {
             "npmName": "@polymer/polymer",
-            "npmVersion": "3.0.5",
-            "jsVersion": "2.6.1"
+            "npmVersion": "3.1.0",
+            "jsVersion": "2.7.0"
```
tomivirkki: IIRC only maintenance bumps in dependencies are allowed for platform patches

web-padawan: I can revert Polymer 2.6.1 but is doesn't make sense to pin 3.x at all
currently 2 versions of Polymer are being installed because of that.

ripla: Unless things are completely broken, we should not do minor or major version bumps in platform maintenance versions.

PR #508, opened 2019-01-18
Update v13 components releases

tomivirkki:


versions.json
```diff
@@ -221,7 +221,7 @@
         "vaadin-text-field": {
             "npmName": "@vaadin/vaadin-text-field",
             "javaVersion": "1.2.1",
-            "jsVersion": "2.1.5",
+            "jsVersion": "2.3.0-alpha7",
```
web-padawan: Note: this will cause dependency duplicates in npm. To prevent it, we need to bump select, combo-box, date-picker and time-picker accordingly.

tomivirkki: Anyway to work around this other than that? Bumping minors of all the dependent components sounds tedious, especially if we need to do it every time some shared dependency enters alpha.

Issue #515, opened 2019-01-24
Specify minimum versions for NPM


ripla:
### Problem
It has turned out that locking dependency versions, e.g. `vaadin-text-field` 1.2.3 for `npm` in a platform release is way too restrictive to be useful in practice. When the text field version is locked to 1.2.3, you will immediately end up with duplicate versions of the component if you add another component which requires a newer (or older) version of text-field. Today this happens e.g. if you try to use the latest `vaadin-login` alpha with Vaadin 12 as `vaadin-login` requires `vaadin-text-field` `2.3.0-alpha3` or newer but Vaadin 12 defines `vaadin-text-field` 2.1.5 should be used. `npm` happily deals with this by installing two versions and the project is broken

### Solution
Starting from Vaadin 13, we will no longer lock the versions for `npm` in the platform. All dependencies are specified as minimum version required, allowing all newer patch and minor versions to be used. We will instead provide a version lock file for the starters so that the application builds are repeatable. We will encourage users to include a lock file in their projects also to avoid surprising dependency updates when a new version of something is released

### Tasks
- [x] Change npm releases to use `^` instead of exact versions
- [x] Create an npm shrinkwrap for `vaadin` and `vaadin-core` when releasing
- [x] Separately release `@vaadin/vaadin`, `@vaadin/vaadin-shrinkwrap`, `@vaadin/vaadin-core` and `@vaadin/vaadin-core-shrinkwrap`
- [x] Update repository readme with information about npm releases
- [x] Update V13 starters to use shrinkwrap dependencies

ripla:
This is a superset of https://github.com/vaadin/platform/issues/495

Artur-:
Does this mean that you would depend on `@vaadin/vaadin-shrinkwrap` instead of `@vaadin/vaadin` if you want it to use specific versions? 

Do those versions defined in a shrinkwrap for `@vaadin/vaadin-shrinkwrap` really affect transitive dependencies also or will it only cause a specific version of e.g `vaadin-button` to be installed inside `node_modules/@vaadin/vaadin/node_ modules`, which is not what you want? Say e.g. that `vaadin-grid` depends on a version range of `vaadin-button`, will it get another version and the whole shrinkwrap becomes useless? @web-padawan

web-padawan:
Follow-up to the question above, summarising the current effort:

1. two packages to be released:
- `@vaadin/vaadin-core-shrinkwrap` will be released for each `@vaadin/vaadin-core` release
[...]

PR #518, opened 2019-01-28
Change npm releases to use ^ instead of exact versions

Artur-:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/518)

### DIFF
--- a/scripts/generator/src/creator.js
+++ b/scripts/generator/src/creator.js
@@ -28,7 +28,7 @@ function createPackageJson(versions, packageJsonTemplate) {
     for (let [name, version] of Object.entries(versions)) {
         if (version.npmName) {
             const npmVersion = version.npmVersion || version.jsVersion;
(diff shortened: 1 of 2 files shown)


ripla:
This is for #515

ripla:
Not sure about this. What if we used the `npmVersion` field instead?

Artur-:
> Not sure about this. What if we used the `npmVersion` field instead?

Wouldn't this mean duplicating every version and just put `^` in front of it?

ripla:
Yes, that's the downside. But it would also allow us to use different versions for Bower and NPM.

Artur-:
> Yes, that's the downside. But it would also allow us to use different versions for Bower and NPM.

I am not sure when you would want that but still `npmVersion` is not going anywhere. This PR just prefixes all versions with `^` in `package.json` so you *can* use the same version for Bower and npm

PR #571, opened 2019-03-01
Update components maintenance versions

web-padawan:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/571)

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -98,7 +98,7 @@
         "vaadin-grid": {
             "npmName": "@vaadin/vaadin-grid",
             "javaVersion": "3.0.2",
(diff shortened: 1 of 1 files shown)


qtdzz:
@web-padawan The build failed because both of the webjar artifacts are not deployed in https://www.webjars.org/

qtdzz:
Also, should we also update `vaadin-text-field:2.3.4` and `vaadin-material-styles:1.2.1` ?

web-padawan:
Thanks. Will take care of the webjars.
Yes, material styles 1.2.1 was incompletely deployed. Will release 1.2.2 first and then update this PR.

PR #586, opened 2019-03-15
Run PR validation on chrome only

### LINKED ISSUE #561: Run IT tests only on Chrome
Investigate if we could speed up PR builds by running the IT tests only in local Chrome.

qtdzz:
Fixes #561

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/586)


qtdzz:
According to @ZheSun88, the current situation is fine. The PR is out-of-date and seems to be not useful anymore. I will close it and leave the branch there, just in case.

PR #657, opened 2019-05-15
Add test for GridContextMenu

### LINKED ISSUE #656: Add test for GridContextMenu
14.0.0.alpha3 contained incompatible versions of `vaadin-grid-flow` and `vaadin-context-menu-flow`.

`GridContextMenu` component should be used in the platform tests to avoid this in the future.

pekam:
fix #656

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/657)


pekam:
The bower mode test is failing as `GridElement::getCell` throws NPE.

Is this also the reason why the Grid test is ignored at the moment?

tulioag:
> The bower mode test is failing as `GridElement::getCell` throws NPE.
> 
> Is this also the reason why the Grid test is ignored at the moment?

Yes, that is the reason. It is failing only when running in in Travis.

pekam:
Should we `@Ignore` this test also and merge the PR, or wait for the issue to be fixed?

tulioag:
I would prefer to wait until the issue is fixed.

pekam:
Ok 👍

PR #877, opened 2019-09-30
added a release notes template for the maintenance release

juhopiirainen:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/877)


juhopiirainen:
Re-organized a bit. Now they (stable and maintenance) should be quite close to each other in style.

PR #888, opened 2019-10-02
Set clientSideMode to be false in platform tests

qtdzz:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/888)


qtdzz:
As @ZheSun88 pointed out, the PR validation build on `master` branch is green even without this change. The reason is that PR validation build uses the defined versions in `versions.json` which is `3.0.0.alpha3` for Flow. This change only affects the SNAPSHOT build because the flipping default mode is still in flow-3.0-SNAPSHOT. I run the tests for SNAPSHOT versions from my local. They are all passed.

ZheSun88:
great. then let us get it in.

PR #941, opened 2019-10-28
Update Flow to 3.0.0.alpha6

qtdzz:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/941)


qtdzz:
There seems to be multiple versions of some transitive dependencies. I will check again in flow artifacts

Issue #944, opened 2019-10-29
Re-enable enforcer plugin in master


qtdzz:
In #941, I removed the enforcer plugin to skip the check temporarily.
After resolving all the dependencies issues in flow and platform, we **should** enable the enforcer plugin again.

haijian-vaadin:
put on hold for now, let's recheck this after the first beta is out.

haijian-vaadin:
Let's close this one. The current assumption is that this plugin was used mostly for the webjars for bower, now they are not needed.

PR #1021, opened 2019-11-26
Re-enable the enforcer plugin

ZheSun88:
As the referred PR in #944 has been released with flow 3.0.0.alpha8
we should re-enable the enforce plugin

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/1021)


ZheSun88:
the issue hasn't been fixed fully. so this PR can be closed.

PR #1086, opened 2020-01-02
Mention pnpm update in 14.2 release note

ZheSun88:
This probably a temporary version. will need to revisit the whole content before beta/rc releases

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/1086)


juhopiirainen:
This needs to go to master branch also?

ZheSun88:
actually, I have thought about this and checked the release note for flow 3.0, where they didn't mention pnpm. so probably, when V15 gets released, V14.2 has been published already(?), if so, pnpm is not a new feature anymore, so probably not needed.

juhopiirainen:
Ok, sounds reasonable. Let's keep it this way at the moment.

PR #1096, opened 2020-01-09
Update browser support release notes for V15

pleku:
Removed IE11 and adjusted Safari support to version 13.

Part of vaadin/flow#6856

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/1096)


pleku:
Not sure if these should be applied to other templates too.

PR #1313, opened 2020-04-27
Enable vaadin-prerelease for snapshot

ZheSun88:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/1313)

### DIFF
--- a/vaadin-spring-boot-starter/pom.xml
+++ b/vaadin-spring-boot-starter/pom.xml
@@ -27,16 +27,6 @@
             <id>vaadin-prereleases</id>
             <url>https://maven.vaadin.com/vaadin-prereleases</url>
         </repository>
-        <repository>
-            <id>vaadin-snapshots</id>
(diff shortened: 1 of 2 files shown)


vaadin-spring-boot-starter/pom.xml
```diff
@@ -27,16 +27,6 @@
             <id>vaadin-prereleases</id>
             <url>https://maven.vaadin.com/vaadin-prereleases</url>
         </repository>
-        <repository>
```
juhopiirainen, line 30: Should the https://maven.vaadin.com/vaadin-prereleases have:

<snapshots>
                <enabled>true</enabled>
</snapshots>
?

ZheSun88: by default, snapshots are set to true.
but yeah.. probably have the snapshot setting is more clear. will add it.

Issue #1349, opened 2020-05-19
Scan the full platform build for known vulnerabilities


Legioth:
Scan the full transitive dependency tree of the whole Vaadin platform for known vulnerabilities as part of each platform release. In this way, we would notice any cases when we're accidentally introducing transitive dependencies to third-party library versions with known dependencies.

For the Maven parts, https://search.maven.org/artifact/org.owasp/dependency-check-maven/5.3.2/maven-plugin could be used. For the npm parts, the built-in `npm audit` command could be used.

ZheSun88:
Add first try to integrate the plugin into platform. 
https://github.com/vaadin/platform/pull/1350

manolo:
Closing so as we can track the same in #1386

PR #1350, opened 2020-05-19
[WIP] add dependency check maven plugin

ZheSun88:
related to #1349 

This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/1350)


pom.xml
```diff
@@ -145,6 +145,26 @@
                 </plugin>
             </plugins>
         </pluginManagement>
-    </build>
 
+        <plugins>
```
manolo, line 149: Apparently it's gonna be executed in all builds, I wouldn't run this always but in CI before beta or rc releases it is enough to run the goal `org.owasp:dependency-check-maven:5.2.2:check` from command line

ZheSun88: Currently, this is only a test for collecting ideas. 
The plan is adding this as a profile, then do the check for the necessary steps. 
running this before beta or rc release is fine, as long as they are not too late. I'd prefer to move the checking even earlier. 

also, I am planning to make use of 
```
<configuration>
        <failBuildOnCVSS>8</failBuildOnCVSS>
</configuration>
```
to limit the vulnerabilities, as there might be some false positive reports.

TatuLund: Just for information, it is also possible to exclude test and provided scope dependencies from the check if that is desired.

ZheSun88: oh, i will check the result after excluding the test and provided dependencies.

ZheSun88: difference between skipping the test/provided dependencies is trivial. 
one solution here can be that we pin those dependencies in platform level. 
introduce a `<dependencyManagement>` in the root, then declare those in submodules. 

a proper way would be fix those from components

ZheSun88:
Finally [this issue](https://github.com/jeremylong/DependencyCheck/issues/2651) has been fixed. so we need to update plugin version to 6.0.3 and test

PR #1380, opened 2020-06-05
Add java LC to versions.js and bom

### LINKED ISSUE #1363: add java LC to BOM
LC is used by MPR and TB and it would help both if version could be managed by bom and avoid extra releases or having an old version.

alvarezguille:
Fixes #1363

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/1380)


alvarezguille:
@manolo this should the cherry picked to LTS

PR #1415, opened 2020-06-30
Created the Acceptance Criteria (AC) template

martaolszewskatusiek:
Created a proposal for the template for the Acceptance Criteria checklist, where there is a user story followed by a checklist.
The checklist contains the case-specific criteria, and then also the generic ones (for APIs, performance, UX/DX tests and design).

The part under construction is meant for replacing the User Value document in the second iteration of AC changes.
The AC template is designated by default title AC, and a label `acceptance criteria`.

The template is to be used in epics or in single issues.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/1415)

PR #1424, opened 2020-07-02
Add Bug report template

ZheSun88:
https://github.com/vaadin/platform/pull/1415#issuecomment-652614880

This change is [Reviewable](https://reviewable.io/reviews/vaadin/platform/1424)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,33 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: "[BUG]"
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: "[BUG]"
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
```
manolo, line 10: wow, you are fast! I was thinking on do the same :-) 

Anyways, I was thinking to align this template with flow and components somehow 
https://raw.githubusercontent.com/vaadin/flow/master/.github/ISSUE_TEMPLATE.md
https://github.com/vaadin/vaadin-grid/issues/new

Another options are:
option 1. in this template we should encourage ppl to open the issue in the appropriate repo if they can, e.g flow or appropriate component
option 2. have more templates for generic issues, component issues, flow issues, then we can move those to the appropriate repo

ZheSun88: that is probably good. i think the main content of those templates is quite similar. (you are free to change the contents here too.)

we can mention those to encourage ppl. but TBH, normally only the new users will use our repo for bugs, experienced users will find the correct one.

PR #1691, opened 2020-11-27
feat: allow specify component version when generating poms

### LINKED ISSUE #1689: Update the bom generate scrips for enabling using different version between component and platform
for example, releasing platform rc with component final version.

manolo:
fixes #1689


manolo:
closing it because it's not the best idea

PR #1791, opened 2021-01-21
Bump CE to 3.0.0 stable

pekam:


ZheSun88:
Is there something we should test on platform level? a smoke test is enough.

pekam:
I thought that the platform team is writing a smoke test for CE..? But if that's a misunderstanding, we can of course create one.

ZheSun88:
we are indeed starting to check the missing tests for modules in platform, so if it is not big burden to add it, can you help?

pekam:
Yes. :+1:  I've created a ticket about it for myself in our backlog and will probably work on that on Monday.

I guess it shouldn't block this PR though? CE is already there as a dependency and this PR is just bumping the version.

manolo (approved):

PR #1794, opened 2021-01-25
Add smoke test for CollaborationAvatarGroup

pekam:
By attaching two groups with distinct users, and asserting
that two avatars are visible, it ensures that Collaboration
Engine is working and synchronizing the avatars.

### DIFF
--- a/vaadin-platform-test/src/main/java/com/vaadin/platform/test/ComponentsView.java
+++ b/vaadin-platform-test/src/main/java/com/vaadin/platform/test/ComponentsView.java
@@ -26,8 +26,8 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
-import org.apache.commons.io.IOUtils;
-
+import com.vaadin.collaborationengine.CollaborationAvatarGroup;
+import com.vaadin.collaborationengine.UserInfo;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.HtmlComponent;
 import com.vaadin.flow.component.accordion.Accordion;
@@ -140,6 +140,8 @@ import com.vaadin.flow.data.provider.hierarchy.HierarchicalQuery;
 import com.vaadin.flow.data.value.ValueChangeMode;
 import com.vaadin.flow.internal.MessageDigestUtil;
 import com.vaadin.flow.router.Route;
+
+import org.apache.commons.io.IOUtils;
 
 @Route("")
 public class ComponentsView extends AppLayout {
@@ -552,9 +554,15 @@ public class ComponentsView extends AppLayout {
 
         AccordionPanel accordionPanel = new AccordionPanel("AccordionPanel", new Span("Content"));
 
+        // Test for avatar components
         Avatar avatar = new Avatar("Donald");
         AvatarGroup avatarGroup = new AvatarGroup(new AvatarGroupItem("Pluto"), new AvatarGroupItem("Mickey"));
 
+        // Tests for collaboration engine
+        CollaborationAvatarGroup collaborationAvatarGroup =
+                new CollaborationAvatarGroup(new UserInfo("foo", "foo"), "topic-id");
+        CollaborationAvatarGroup collaborationAvatarGroup2 =
+                new CollaborationAvatarGroup(new UserInfo("bar", "bar"), "topic-id");
 
         // These components are flow internal classes, these lines is to make pass the ComponentUsageTest
         JavaScriptBootstrapUI javaScriptBootstrapUI;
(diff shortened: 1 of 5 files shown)


ZheSun88:
i put a PR to update the component side.. if it works. https://github.com/vaadin/vaadin-flow-components/pull/593

pekam:
Thanks! But I think something else is still broken here. Adding the explicit dependency to `collaboration-engine` should've fixed the version conflict (I just tested with it, should be removed), but the build is still failing and I can't see any relevant information in the logs. I can't also run this locally in production for some reason. `mvn jetty:run -Pproduction` doesn't seem to work: `Vaadin is running in DEBUG MODE`.

vaadin-platform-test/collaborationEngineDataDir/ce-license.json
```diff
@@ -0,0 +1 @@
+{"content":{"key":"789f6a8e-a5f3-4eba-80e5-7f062f68ccb0","owner":"integration tests","quota":1000000,"endDate":"2500-12-31"},"checksum":"OizVdCFUmbzkzzXsHc35u9lL2m7vYKCX6gONe3cCRds="}
```
manolo, line 1: Why do we need this?
Is this a valid license, I mean can I copy and paste in my Flow app with CE and it works?

vaadin-platform-test/src/main/java/com/vaadin/platform/test/ComponentsView.java
```diff
@@ -390,6 +392,13 @@ protected void setPresentationValue(String newPresentationValue) {
         accordion.add("Title", new Paragraph("Content"));
         accordion.setId("accordion");
 
+        CollaborationAvatarGroup collaborationAvatarGroup =
```
manolo, line 395: Is this actually testing the CE? I mean I don't see any binding with an actual field

vaadin-platform-test/src/test/java/com/vaadin/platform/test/ChromeComponentsIT.java
```diff
@@ -547,6 +548,17 @@ public void contextMenuIsRenderedAndRecievesItemSelectionEvents() {
...
 
+    @Test
+    public void collaborationAvatarGroupIsRendered() {
+        AvatarGroupElement group = $(AvatarGroupElement.class)
+                .id("collaborationavatargroup");
+        assertElementRendered(group);
+        Assert.assertEquals("foo", group.getAvatarElement(0)
+                .getPropertyString("name"));
+        Assert.assertEquals("bar", group.getAvatarElement(1)
+                .getPropertyString("name"));
```
manolo, line 559: Shouldn't we test that changing an input something appears somewhere indicating that the CE is working and the field is being modified?

manolo (changes requested):

vaadin-platform-test/collaborationEngineDataDir/ce-license.json
pekam: Yes, it's a valid license, and technically you can just copy it to your Flow app and CE will work (but sharing one license across apps would be violating the license terms).

We need it here because the tests are running in production mode. CE licensing works like so that in dev mode you can use it for free, but for prod mode, you need to obtain this kind of license from Vaadin that defines the `quota` (how many end users are allowed) and `endDate`.

vaadin-platform-test/src/main/java/com/vaadin/platform/test/ComponentsView.java
pekam: When we attach `CollaborationAvatarGroup`, it updates data in CE to include a new avatar, which will update all the other `CollaborationAvatarGroups` to include that avatar. So, by attaching two avatar groups with the same topic id, and checking that they contain two avatars, we make sure that data syncing with CE works.

We could have also created two fields connected to separate `CollaborationBinders`, and make sure that field values are synced.

Even better would be to open two UIs and make sure that data is synced between those, but that needs lots of extra setup that I wanted to avoid here.

But I was thinking that having one feature that synchronizes data through CE in the same UI would be enough for a smoke test.

manolo: Gotcha, it's alright then.

vaadin-platform-test/src/test/java/com/vaadin/platform/test/ChromeComponentsIT.java
manolo: Discard my comment, I see now the approach

vaadin-platform-test/collaborationEngineDataDir/ce-license.json
manolo: Ok, clear enough. Now my question is can we provide this somehow using deployment vars? like env vars or system properties ?  This repo is public and I don't think it's a good idea to keep this here

manolo (approved):

PR #1837, opened 2021-02-05
core: update mpr version to 5.0.0.beta1 in versions.json

miguelatvaadin:


ZheSun88 (commented):
Thanks for your contribution, you are going to make Vaadin Platform better.

PR #1975, opened 2021-03-29
chore: fix versions for web-components links

manolo:


ZheSun88:
there are a few more places using the `vaadin-` prefix, such as https://github.com/vaadin/platform/blob/master/scripts/generator/src/creator.js#L153, https://github.com/vaadin/platform/blob/master/scripts/generator/templates/template-release-notes.md

PR #1991, opened 2021-04-06
feat: add support for publishing vaadin-gradle-plugin

taefi:
Similar to vaadin-maven-plugin be based on flow-maven-plugin, 
publishing vaadin-gradle-plugin is based on flow-gradle-plugin.
With every flow release there would a flow-gradle-plugin which 
platform would repackage its sources and would publish 
vaadin-gradle-plugin with same version as vaadin-platform.
As vaadin gradle plugin is set up to be built and published using 
gradle, there would be an extra step in releasing final versions of 
the platform for calling *publishPlugins* task from *build.gradle* 
(inside vaadin-gradle-plugin module) to publish an official gradle 
plugin to gradle plugin central.


ZheSun88:
PR is green.. now we will wait for a released flow version.

scripts/generator/templates/template-vaadin-spring-bom.xml
```diff
@@ -90,12 +90,12 @@
             <dependency>
                 <groupId>com.vaadin</groupId>
                 <artifactId>vaadin-lumo-theme</artifactId>
-                <version>${project.version}</version>
+                <version>${vaadin.material.theme.version}</version>
```
knoobie, line 93: This looks odd

ZheSun88: nice catch.. i will update it. Thanks

taefi: Thanks, @Knoobie! I'll ask @ZheSun88 to step in...

versions.json
```diff
@@ -1,7 +1,7 @@
 {
     "core": {
         "flow": {
-            "javaVersion": "7.0.0.alpha3"
+            "javaVersion": "7.0-SNAPSHOT"
```
taefi, line 4:
```suggestion
            "javaVersion": "7.0.0.alpha4"
```

ZheSun88 (approved):

PR #2054, opened 2021-04-26
chore: replace version with the appropiate value in vaadin_versions.json

manolo:


pleku (approved):
Tested locally and it works

PR #2135, opened 2021-05-26
chore: add vaadin-dev-server dependency

joheriks:
The dev server and live reload functionality has recently been
factored out of flow-server (so that it can be excluded in 
production mode). Add the dependency to the new module
(vaadin-dev-server) to vaadin-core to not break existing projects.

### DIFF
--- a/vaadin-core/pom.xml
+++ b/vaadin-core/pom.xml
@@ -44,6 +44,10 @@
         </dependency>
         <dependency>
             <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-dev-server</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
             <artifactId>flow-lit-template</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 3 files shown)


vaadin-core/pom.xml
```diff
@@ -42,6 +42,10 @@
             <groupId>com.vaadin</groupId>
             <artifactId>fusion-endpoint</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-dev-server</artifactId>
```
ZheSun88, line 47: is this artifact available now ?
didnt find it in the snapshot repo http://tools.vaadin.com/nexus/content/repositories/vaadin-prereleases/com/vaadin/

ZheSun88: oh, still running https://bender.vaadin.com/viewLog.html?buildId=233042&tab=buildResultsDiv&buildTypeId=Flow_Flow80_Flow80Snapshot

ZheSun88:
Add the dep to `spring-bom` with flow version. 

as this is the new artifact which hasn't been included in any flow releases, i updated the flow version to `8.0-SNAPSHOT`

ZheSun88 (approved):

PR #2177, opened 2021-06-07
feat: add virtual list component

tomivirkki:


tomivirkki:
@ZheSun88 Seems that the missing test case isn't related to this PR

PR #2214, opened 2021-06-22
fix: remove vaadin-bom from maven plugin build

pleku:
The used dependency versions for Maven plugin should be the same as what are used in the flow build.
There was an issue now that the plugin is getting an outdated commons-io version from the components.

Follow up step is to fix/pin the commons version for `vaadin-bom` to a newer one.


ZheSun88:
looks the WC update breaks the hybrid module, i will combine  this fix with the component version update PR. https://github.com/vaadin/platform/pull/2206

PR #2246, opened 2021-07-09
bump buddy version in bnd-tools-tests [skip ci]

manolo:


ZheSun88:
we can merge it in for testing, but eventually we should probably remove this dependency.

ZheSun88 (approved):

PR #2530, opened 2021-11-09
chore: update Safari and iOS versions for V22

web-padawan:
## Description

In Vaadin 22 we use some APIs like [gap property for Flexbox](https://caniuse.com/flexbox-gap) that require Safari 14.1 / iOS 14.5+

## Type of change

- Release notes update


jouni:
Is this template also used for Vaadin 14.x?

web-padawan:
@jouni The 14.8 templates are here: https://github.com/vaadin/platform/tree/14.8/scripts/generator/templates

PR #2609, opened 2021-12-16
feat: add hilla bom and artifact

haijian-vaadin:
add hilla bom and artifact, so that a user project can depend on hilla instead of vaadin.


hilla/src/main/templates/dev/hilla/shrinkwrap/VaadinCoreShrinkWrap.java
```diff
@@ -0,0 +1,25 @@
...
+ * the License.
+ */
+package dev.hilla.shrinkwrap;
+
+import com.vaadin.flow.component.dependency.NpmPackage;
+
+/**
+ * Empty class which makes flow add npm vaadin-core-shrinkwrap dependency.
+ */
+@NpmPackage(value = "@vaadin/vaadin-core-shrinkwrap", version = "${shrink.version}")
```
manolo, line 23: I think we can add here `@NpmPackage(value="@vaadin/vaadin-core")` and package.json file will be updated with all the components.

haijian-vaadin: Per discussion, removed this file all together

hilla/pom.xml
```diff
@@ -0,0 +1,173 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-dev-server</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-client</artifactId>
+        </dependency>
+        <dependency>
```
manolo, line 49: We must remove lumo and material dependency here.
A pure TS project might want not to use vaadin-components nor vaadin-themes 

But previously we need to fix https://github.com/vaadin/flow/issues/12606

manolo (changes requested):

manolo:
Before merging this, we should fix https://github.com/vaadin/flow/issues/12606

hilla/pom.xml
haijian-vaadin: hmm, I think it's fine to have those dependencies for now. As the main goal is to have a `hilla-bom` and dependency that a user can use.
Separate Hilla from Flow is the ultimate goal, but not necessarily to have it now. For example, `fusion-endpoint` depends on `flow-server` anyway.

manolo (approved):

hilla/pom.xml
manolo: ok, let lave it for now

ZheSun88 (approved):

PR #2644, opened 2022-01-10
feat: update compilation target to 11

caalador:
As all dependents are build with jdk11 we need to update the platform to also build with target version 11


ZheSun88:
Thanks @caalador

ZheSun88:
there is another PR #2645  which contains more sufficient changes.. i will close this one

PR #2646, opened 2022-01-10
chore: refactor hilla artifacts

ZheSun88:
as hilla bom will be released from version 1.0. extracting `hilla.version` as a property, so that we can set a version before doing a release


ZheSun88:
need to add one step to set the `hilla.version` before release.

manolo (approved):

PR #2726, opened 2022-02-08
chore: add map component to platform

sissbruecker:
## Description

Adds the map component to the platform.

### DIFF
--- a/vaadin-platform-test/src/main/java/com/vaadin/platform/test/ComponentsView.java
+++ b/vaadin-platform-test/src/main/java/com/vaadin/platform/test/ComponentsView.java
@@ -595,6 +595,8 @@ public class ComponentsView extends AppLayout {
         MessageInput messageInput = new MessageInput();
         messageInput.addSubmitListener(e -> log.log(e.getValue()));
 
+        com.vaadin.flow.component.map.Map mapComponent = new com.vaadin.flow.component.map.Map();
+
         // These components are flow internal classes, these lines is to make pass the ComponentUsageTest
         JavaScriptBootstrapUI javaScriptBootstrapUI;
         WebComponentUI webComponentUI;
@@ -642,6 +644,7 @@ public class ComponentsView extends AppLayout {
         components.add(messageList);
         components.add(messageInput);
         components.add(main);
+        components.add(mapComponent);
 
         layouts.add(formLayout);
         layouts.add(verticalLayout);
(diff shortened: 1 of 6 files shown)


ZheSun88:
I added the `vaadin-map` entry to **versions.json** too.. this is because we are still using the older way in platform for vaadin-component. 

have talked with Serhii about this, and we will probably remove it in later V23 minors. 


also, i guess the build will fail on the test coverage as this component has not been added to the test classes. let us see.

ZheSun88:
I will fix the test for this PR. 

but i think we will not merge it before upgrading the project to V23.1-SNAPSHOT. As this feature is not part of 23.0, adding it to the vaadin-bom will make it totally public to the end users. 

if you need to use this compoment, you will still need to manually add the related dependency to your project.

Artur-:
Why is this feature flagged component not part of 23? That would make it harder to get any feedback on it before removing the flag

jouni:
Came here to say the same as Artur, that we want this in 23.0. This is different from how we’ve used to do things, but that’s changing, that experimental features / feature previews can be part of any platform release.

vaadin-platform-test/src/main/java/com/vaadin/platform/test/ComponentsView.java
```diff
@@ -595,6 +595,8 @@ protected void setPresentationValue(String newPresentationValue) {
         MessageInput messageInput = new MessageInput();
         messageInput.addSubmitListener(e -> log.log(e.getValue()));
 
+        com.vaadin.flow.component.map.Map mapComponent = new com.vaadin.flow.component.map.Map();
```
manolo, line 598: We should not have smoke tests for experimental components in the platform until they go to production

vaadin-platform-test/src/main/resources/vaadin-featureflags.properties
```diff
@@ -0,0 +1,2 @@
+# Map component (Pro)
+com.vaadin.experimental.mapComponent=true
```
manolo, line 2: I'm doing some modifications in flow so as we don't need to enable experimental components in our test suite.
Blocking this for now

manolo (changes requested):

vaadin-platform-test/src/main/resources/vaadin-featureflags.properties
manolo: ok lets merge this, and fix flow later

manolo (approved):

PR #2738, opened 2022-02-15
chore: generate the bundle package.json

ZheSun88:


manolo (approved):
LGTM, please @platosha take a look

platosha:
- After running `sh scripts/generateBoms.sh`, I’ve got `version: "23.0-SNAPSHOT"` in the `bundles/package.json`. npm cannot release a version like that, I think.
- `bundles/package.json` has dependency to `@vaadin/vaadin`, which needs to be in sync with the bundles package version, but is not updated by the script.

ZheSun88:
- for releasing, we first set the project version to platform, then run the scripts/generateboms.sh , in that case, the version will be overwritten with the released version.

- add the script to update the `@vaadin/vaadin` version

Updated the branch based on the comments. @platosha

platosha (approved):

PR #2756, opened 2022-02-22
chore: remove no longer used iron-a11y-announcer

web-padawan:
## Description

Starting from [23.0.0-alpha3](https://github.com/vaadin/web-components/releases/tag/v23.0.0-alpha3), web components use internal logic instead of the legacy `iron-a11y-announcer` dependency.

## Type of change

- Dependency update

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -109,11 +109,6 @@
             "jsVersion": "23.0.0-beta3",
             "npmName": "@vaadin/integer-field"
         },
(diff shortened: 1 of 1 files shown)


web-padawan:
The rest of `iron-` dependencies comes from `iron-list` and `iron-icon` which we still use in Flow components.

ZheSun88 (approved):

PR #2772, opened 2022-02-24
chore: update @vaadin/bundles to 23.0.0-rc1

platosha:
Depends on #2767


ZheSun88:
will combine this PR with #2767 to reduce the waiting time.

PR #2849, opened 2022-03-23
chore: remove flow-spring entry from version.json

ZheSun88:
since Vaadin 23, spring addon has been moved to flow project, so let us remove the separate entry for spring.


ZheSun88:
@knoobie  no worries.. in Vaadin 23.0.5, there is only component updates.

PR #2898, opened 2022-04-05
Classic Components are not part of the Design System [skip ci]

jouni:
Classic Components are more related to MPR.


ZheSun88 (approved):

jouni:
Thanks @ZheSun88

PR #2900, opened 2022-04-05
Add Classic Components to prerelease release notes [skip ci]

jouni:


jouni:
These changes should probably be picked to `master` as well, right?

ZheSun88 (approved):

ZheSun88:
i will combine this with #2898

PR #3182, opened 2022-08-11
chore: do not run chrome tests in sauce

manolo:


vaadin-platform-test/src/test/java/com/vaadin/platform/test/ComponentsIT.java
```diff
@@ -26,7 +26,7 @@ public class ComponentsIT extends AbstractPlatformTest {
         String browsers = System.getProperty("grid.browsers");
         if (sauceUser != null && !sauceUser.isEmpty()) {
             if (browsers == null || browsers.isEmpty()) {
-                Parameters.setGridBrowsers("firefox,chrome,safari,edge");
+                Parameters.setGridBrowsers("firefox,safari,edge");
```
ZheSun88, line 29: curious: why remove chrome from test?

alvarezguille (approved):

vaadin-platform-test/src/test/java/com/vaadin/platform/test/ComponentsIT.java
manolo: Because we run twice this class, this list is only when running in sauce, when the class is run in local or in hub it's always chrome

PR #3186, opened 2022-08-12
chore: Upgrade to TestBench 8.1 beta2

Artur-:


Artur-:
It should not be possible for beta2 to break anything. It only adds the possibility to set a timeout

manolo:
@Artur- as you can see in these [builds](https://teamcity.vaadin.com/buildConfiguration/VaadinPlatform_VaadinPlatformPrValidation?branch=3186&buildTypeTab=overview&mode=builds) after setting beta1 it goes green

It also happens in platform [snapshots](https://bender.vaadin.com/buildConfiguration/VaadinPlatform_TestReleasePlatformSnapshotToMavenVaadinPrerelease?branch=%3Cdefault%3E&mode=builds)  it went green when I temporary deployed a snapshot in pre-releases with the beta1 hash

For some reason the capabilities that are passed to the hub do not match any browser, we thought at first that it was problem in selenoid not having resources for more browsers.

BTW, platform now is tested in sauce + hub after the [splitting](https://github.com/vaadin/platform/pull/3182)  tests in two failsafe executions

ZheSun88:
close this, as #3189 has been made

PR #3363, opened 2022-10-20
include security requirements in AC template [skip ci]

manolo:


Legioth:
Based on Slack discussion, there's simplified suggestion to have only a single question like this:
> [ ] Security implications have been taken into account (elaborate or link to product security requirement specification if there are implications)

alvarezguille (approved):

PR #3364, opened 2022-10-20
chore: Update license checker to 1.10.1

mshabarov:


ZheSun88:
with Vaadin 14, we are still support the webjars, the jsVersion there is used for the 
```
            <dependency>
                <groupId>org.webjars.bowergithub.vaadin</groupId>
                <artifactId>license-checker</artifactId>
                <version>{{vaadin.vaadin-license-checker.jsVersion}}</version>
            </dependency>
```

what should do with this? remove it, does the license-checker work under bower-mode?

mshabarov:
> with Vaadin 14, we are still support the webjars, the jsVersion there is used for the
> 
> ```
>             <dependency>
>                 <groupId>org.webjars.bowergithub.vaadin</groupId>
>                 <artifactId>license-checker</artifactId>
>                 <version>{{vaadin.vaadin-license-checker.jsVersion}}</version>
>             </dependency>
> ```
> 
> what should do with this? remove it, does the license-checker work under bower-mode?

Good catch! Not tested yet, but JS side license checker is independent from new license checker, because it directly calls license server https://github.com/vaadin/license-checker/blob/8f27627b957a0a7064ac0849a4f9ebddd1371e27/src/vaadin-license-checker.js#L252 . So should be fine, but we need to test for sure.

PR #3387, opened 2022-10-26
chore: remove sauce-connect-plugin (14.8)

ZheSun88:


ZheSun88:
V14.8 will be deprecated after V14.9 got released. so we can ignore this updates.

PR #3410, opened 2022-10-28
chore: Exclude not ready dependencies

mshabarov:
Porlet - not planned
OSGi - not planned until we have Felix Jetty 8
MPR - planned for next platform alpha


ZheSun88:
I guess we need to remove some testing modules from platform, because of your change.

ZheSun88 (approved):

PR #3424, opened 2022-11-02
chore: dont overwrite license-checker version to snapshot

ZheSun88:


scripts/generator/src/transformer.js
```diff
@@ -27,10 +27,13 @@ function transformPlatformVersion(versions, platformVersion) {
 
 function transformJavaSnapshots(versions) {
     const majorMinorVersions = /(\d*?\.\d*).*/;
+    let module;
     const snapshotVersionVisitor = (key, value, parent) => {
-        if (key === 'javaVersion') {
+        // dont update  the license-checker version to snapshot
+        if (key === 'javaVersion' && module !== 'vaadin-license-checker') {
```
mshabarov, line 33: how can `module` get the value `vaadin-license-checker` if it always assigned to `key` ?

ZheSun88: this is an iteration. key is basically changing all rounds. 
the module name comes one round before `javaVersion`, so we store it in the module parameter. so when the iteration goes to vaadin-license-checker. we skip the snapshot replacement

mshabarov (approved):

PR #3431, opened 2022-11-03
chore: update testbench to 9.0.0.alpha3

MarcinVaadin:
update dependencies


ZheSun88:
let us combine this with #3429, so that we dont need to wait for the validation twice.

PR #3456, opened 2022-11-10
chore: Include license-checker in bom

Artur-:


Artur-:
This shouldn't be needed as the version is in vaadin-spring-bom

PR #3468, opened 2022-11-16
feat: add hilla-react module to 23.3

cromoteca:
Creates a new module: `dev.hilla:hilla-react`. It will be used in Hilla+React applications instead of `dev.hilla:hilla` to avoid import Vaadin Components directly in `package.json`.

This is related to vaadin/start#2108.

Please review the generated dependencies carefully.


hilla-react/versions.json
```diff
@@ -0,0 +1,329 @@
+{
```
ZheSun88, line 1: does hilla have some logic to parse this file?
for example: 
- replacing the {{version}} to a version number
- keep the version inside to be updated

versions.json
```diff
@@ -5,6 +5,12 @@
             "npmName": "@vaadin/bundles"
         }
     },
+    "react": {
+        "react-components": {
+            "jsVersion": "1.3.0-alpha6",
+            "npmName": "@hilla/react-components"
+        }
+    },
```
cromoteca, line 13: Is adding this section safe?

ZheSun88: It is fine to add it here. This file will not be deployed anywhere.

hilla-react/pom.xml
```diff
@@ -0,0 +1,181 @@
...
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>dev.hilla</groupId>
+            <artifactId>endpoint</artifactId>
+            <version>${hilla.version}</version>
```
ZheSun88, line 40: you can remove this `${hilla.version}` here, as the version should come from the hilla-bom. 
same with the `engine-runtime` artifacts.

ZheSun88:
Is there a way to test this? 
or have you tested this locally?

hilla-react/pom.xml
cromoteca: I'll add `endpoint` to the bom.

cromoteca:
> Is there a way to test this? or have you tested this locally?

Good question. I tested locally and it's fine (together with https://github.com/vaadin/start/pull/2108). But of course something more robust is needed. Is there anything I can add on TeamCity maybe?

manolo:
There is a module that is activated with the `fusion-hybrid` profile that I think is skipped since some time ago

manolo (approved):

platosha (approved):

PR #3481, opened 2022-11-18
pass credentials to the pit script (23.3)

manolo:


ZheSun88:
should we use java 11 for testing? 

```
      - uses: actions/setup-java@v3
        with:
          java-version: '17'
          distribution: 'temurin'
```

manolo:
No, jdk17 is ok since it can compile projects with 11 and we can run PiT as well in 24.0.
Since we are testing JDKs 11 in other CI builds and here we need to know only whether  projects are broken, it's ok to use 17 here

ZheSun88 (approved):

PR #3482, opened 2022-11-18
chore: Add MPR-V8 dependency again

mshabarov:
Note: MPR-V7 not ready yet, so not added.

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -165,6 +165,9 @@
             "jsVersion": "24.0.0-alpha4",
             "npmName": "@vaadin/message-list"
         },
+        "mpr-v8": {
+            "javaVersion": "7.0.0.alpha1"
+        },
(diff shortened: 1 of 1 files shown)


versions.json
```diff
@@ -165,6 +165,9 @@
             "jsVersion": "24.0.0-alpha4",
             "npmName": "@vaadin/message-list"
         },
+        "mpr-v8": {
```
ZheSun88, line 168: do we also need that `mpr-v7`? seems we have both in the spring-bom
https://github.com/vaadin/platform/blob/master/scripts/generator/templates/template-vaadin-spring-bom.xml#L555...L564

ZheSun88: oh.. now i noticed your Note..

ZheSun88 (approved):

PR #3536, opened 2022-11-29
Build default dev bundle

Artur-:


Artur-:
How is vaadin-platform-hybrid-test/frontend/views/components/components-view.ts related to this @ZheSun88 ?

ZheSun88:
because the build was complaining about the missing component import in the component-view.ts

https://teamcity.vaadin.com/viewLog.html?buildId=160665&buildTypeId=VaadinPlatform_VaadinPlatformPrValidation

vaadin-dev-bundle/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+                    <supportedPackagings>jar</supportedPackagings>
+                    <useProvidedScope>true</useProvidedScope>
+                    <systemProperties>
+                        <vaadin.enableDevServer>false</vaadin.enableDevServer>
+                    </systemProperties>
+                </configuration>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>start</goal>
```
ZheSun88, line 71: seems we need a stop jetty here, or starting this on a different port
otherwise, when platform starts the test, the build will complain about port 8080 is already in use

Artur-: Yeah it should also run stop

PR #3677, opened 2023-01-17
chore: add module for generating the platform SBOM with all dependencies

manolo:


scripts/generator/src/creator.js
```diff
@@ -318,7 +318,8 @@ function compareAndBuildJavaComponentReleaseString(versionName, currentVersion,
     let result = '';
     const currentVersionSemver = toSemVer(currentVersion);
     const previousVersionSemver = toSemVer(previousVersion);
-    if (compareVersions(currentVersionSemver, previousVersionSemver) === 1) {
+    // sometimes we use SNAPSHOTS in versions.json e.g. when waiting for a new alpha/beta with a fix
```
ZheSun88, line 321: what is the idea of this change?

we should not release platform if there are snapshot dependencies.

ZheSun88 (approved):

scripts/generator/src/creator.js
manolo: The problem is that in this case, having the CE 6.0-SNAPSHOT in versions.json `currentVersionSemver=6.0-SNAPSHOT` and that is incorrect in the `compareVersions` library. It's just a fix that was not manifested before

ZheSun88: that function is used for formatting the release note...
which is not needed when there is a snapshot dependency in the versions.json. as there wont be any release based on that

manolo: Still that is a JS error that we need to avoid

manolo: Not sure if it could be fixed somehow else, but I couldn't run this PR to test that attachment push woks, otherwise

ZheSun88 (approved):

scripts/generator/src/creator.js
manolo: No version keeps 6.0-SNAPSHOT except for the `compareVersions` that I'm pretty sure will never return 1, then the workflow in the program continues

PR #3688, opened 2023-01-17
chore: add module for generating the platform SBOM with all dependencies (23.3)

manolo:


vaadin-platform-sbom/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+    <parent>
+        <groupId>com.vaadin</groupId>
+        <artifactId>vaadin-platform-parent</artifactId>
+        <version>23.3-SNAPSHOT</version>
+    </parent>
+    <artifactId>vaadin-platform-sbom</artifactId>
+    <packaging>jar</packaging>
+    <properties>
+        <maven.compiler.source>17</maven.compiler.source>
+        <maven.compiler.target>17</maven.compiler.target>
```
ZheSun88, line 14: should we use java 11 here, as it is for V23.
(probably doesn't block anything. )

ZheSun88 (approved):

PR #3728, opened 2023-01-27
feat: Add the default bundle always

mshabarov:
## Description

Adds the default bundle dependency always to the `vaadin-dev`, so it will be always available in express build mode which is default in Flow now.

Related-to https://github.com/vaadin/flow/issues/15761

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [ ] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [ ] New and existing tests are passing locally with my change.
- [ ] I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change

- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.


mshabarov:
Looks unrelated to this change:
```
Run actions/upload-artifact@v3.1.1
Error: No files were found with the provided path: **/target/bom-vaadin.json
**/target/*-report.json
**/target/tree-*.txt
**/target/dependencies.html. No artifacts will be uploaded.
```

ZheSun88:
what is the idea of this PR? 
it is introducing a circular dependency 

```
[ERROR] [ERROR] The projects in the reactor contain a cyclic reference: Edge between 'Vertex{label='com.vaadin:vaadin-dev-bundle:24.0-SNAPSHOT'}' and 'Vertex{label='com.vaadin:vaadin:24.0-SNAPSHOT'}' introduces to cycle in the graph com.vaadin:vaadin:24.0-SNAPSHOT --> com.vaadin:vaadin-core:24.0-SNAPSHOT --> com.vaadin:vaadin-dev-bundle:24.0-SNAPSHOT --> com.vaadin:vaadin:24.0-SNAPSHOT @ 
```

mshabarov:
> what is the idea of this PR? it is introducing a circular dependency
> 
> ```
> [ERROR] [ERROR] The projects in the reactor contain a cyclic reference: Edge between 'Vertex{label='com.vaadin:vaadin-dev-bundle:24.0-SNAPSHOT'}' and 'Vertex{label='com.vaadin:vaadin:24.0-SNAPSHOT'}' introduces to cycle in the graph com.vaadin:vaadin:24.0-SNAPSHOT --> com.vaadin:vaadin-core:24.0-SNAPSHOT --> com.vaadin:vaadin-dev-bundle:24.0-SNAPSHOT --> com.vaadin:vaadin:24.0-SNAPSHOT @ 
> ```

Yeah, I see. The idea is to add `vaadin-dev-bundle` to the standard list of dependencies. How can I add it to a higher level so the `vaadin` and `vaadin-dev-bundle` would be on the same level and don't raise a circularity error.

ZheSun88:
>How can I add it to a higher level so the vaadin and vaadin-dev-bundle would be on the same level and don't raise a circularity error

i think in this case, we should use flow and component dependencies in `vaadin-dev-bundle` instead of using `vaadin` dependency inside, which means duplicating the dependency list in `vaadin`.

vaadin-core/pom.xml
```diff
@@ -42,6 +42,10 @@
             <groupId>com.vaadin</groupId>
             <artifactId>vaadin-dev-server</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-dev-bundle</artifactId>
+        </dependency>
```
knoobie, line 48: I'm wondering: should this be added to vaadin-core or as transitive dependency to vaadin-dev-server? Otherwise people have to also exclude this dependency if they don't use it (all old users?) from their production build if they don't wanna use it / have it in use.

mshabarov: Having it in `vaadin-dev-server` raises a coupling between platform and flow, since this is a platform's module.
We have a general issue about having dev mode related dependencies in production build, see https://github.com/vaadin/flow/issues/15749 . So the idea is to make `vaadin-dev-server` available asap, because the express build mode is now a default and later try to figure out the automatic mechanism for removing all that not needed for production.

knoobie: I know that issue :) Artur closed my issue about it in favor for it. I totally understand your reasoning! Wanted to mention it at least. It's easy for me and other to also exclude or from the build in the meantime.

scripts/generator/templates/template-dev-bundle-pom.xml
```diff
@@ -39,8 +39,215 @@
...
+            <artifactId>vaadin-context-menu-flow</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-menu-bar-flow</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-board-flow</artifactId>
```
knoobie, line 211: Adding also pro / prime components to the default bundle could be a problem, or? I've seen reports that the availability on the class path alone breaks builds for free users.

vaadin-core/pom.xml
Artur-: https://github.com/vaadin/flow/issues/15749#issuecomment-1406356660

scripts/generator/templates/template-dev-bundle-pom.xml
mshabarov: Don't know if the leaking pro / prime was caused by having dev bundle dependency, but I'll rework this PR to follow what's Artur suggested https://github.com/vaadin/flow/issues/15749#issuecomment-1406356660

Artur-: That would be roughly 8a4930fa5a1a9da2dc4619f6b5914583621cf599

mshabarov: Dev bundle module has Vaadin dependencies as optional:
```
        <dependency>
            <groupId>com.vaadin</groupId>
            <artifactId>vaadin</artifactId>
            <optional>true</optional>
        </dependency>
```
So I suppose unless you add `<artifactId>vaadin</artifactId>` to your pom.xml, the pro/prime componens wouldn't leak.

knoobie: > Don't know if the leaking pro / prime was caused by having dev bundle dependency

No, it wasn't. There were reports in discord about people upgrading to v24 and they had vaadin instead of vaadin-core as dependency, resulting in build errors. So I wanted to make sure a similar problem is raised here.

Artur-:
this needs a rebase now

mshabarov:
> this needs a rebase now

Yes, I did a squash and rebase.

mshabarov:
Tested with the starter project:
1. `vaadin-dev-bundle` is included into the project by default within `vaadin-dev` in `vaadin-core`
2. if `vaadin-core` dependency is only set in the project, the commercial components don't leak into dependency tree
3. having `vaadin` dependency in the project includes them into dependency tree.

caalador (approved):

PR #3729, opened 2023-01-27
Move development dependencies to vaadin-dev/hilla-dev

Artur-:


hilla/pom.xml
```diff
@@ -43,8 +43,8 @@
             <artifactId>engine-runtime</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.vaadin</groupId>
-            <artifactId>vaadin-dev-server</artifactId>
+            <groupId>dev.hilla</groupId>
+            <artifactId>hilla-dev</artifactId>
```
cromoteca, line 47: I think this should also be changed in `hilla-react`

Artur-: Yes, thanks

vaadin-core/pom.xml
```diff
@@ -40,7 +40,7 @@
         </dependency>
         <dependency>
             <groupId>com.vaadin</groupId>
-            <artifactId>vaadin-dev-server</artifactId>
+            <artifactId>vaadin-dev</artifactId>
```
mshabarov, line 43: With this modules structure, we would need to remove `vaadin` from `vaadin-dev-bundle` (which is now set as an optional) and replace it with the entire list of components dependencies. Do you think it would be enough to add an optional scope to all of the component in this list to avoid leaking commertial component if the dev bundle dep is added? Asking because this would be a next step and good to figure out the structure in adnvance :)

mshabarov: No, I mean adding optional to the dependencies in the vaadin-dev-bundle.

ZheSun88: this structure looks good to me.. did i miss something?

mshabarov (approved):
With this module

PR #3730, opened 2023-01-27
chore: Simplify production mode specification

Artur-:
the productionMode parameter has not been needed since https://github.com/vaadin/flow/pull/15390


ZheSun88:
oh, so the `flow-server-production-mode` is not needed any more? 
any related changes in flow that can be provided as a commit discription?

Artur-:
flow-server-production-mode has not been needed since... I don't know when

ZheSun88 (approved):

PR #3764, opened 2023-02-07
fix: Exclude BundleComponents class from JAR

mshabarov:
So it would add the AtNpmPackage dependencies into a dev bundle, but wouldn't spread them to the projects.


scripts/generator/templates/template-dev-bundle-pom.xml
```diff
@@ -279,6 +279,7 @@
                 <configuration>
                     <excludes>
                         <exclude>**/*Fake*</exclude>
```
Artur-, line 281: Can this "Fake" be removed now? I think it was for the same purpose when the class was named differently

Artur-: Oh, there is also a FakeAppConf. Why aren't the annotations on that class?

Artur- (approved):

Artur- (approved):

scripts/generator/templates/template-dev-bundle-pom.xml
mshabarov: Done

PR #3772, opened 2023-02-08
test: Test that a fresh starter project does not need a bundle build

mshabarov:
Fixes https://github.com/vaadin/flow/issues/15862


mshabarov:
Tested manually with a starter project, but need to add a test for checking that the default bundle is actually used.

vaadin-dev-bundle/src/main/java/com/vaadin/devbundle/FakeAppConf.java
```diff
@@ -9,6 +9,14 @@
 @JsModule("@vaadin-component-factory/vcf-nav")
 @NpmPackage(value = "line-awesome", version = "1.3.0")
 @NpmPackage(value = "@vaadin-component-factory/vcf-nav", version = "1.0.6")
+@NpmPackage(value = "@vaadin/bundles", version = "%bundles.version%")
```
Artur-, line 12: Why do we need these?

mshabarov: Just because they are missing in the dev bundle :)
To be honest I'm not much familiar with the `@vaadin/bundles`.
What I understand is that this package was included into dev bundle because there was a dependency to `vaadin` which imports it.
Now we change the structure of dependencies in platform and it's not available for bundle maker.
Why are the rest of dependencies are missing I have no idea.

mshabarov:
@ZheSun88 I've added a test module to the root which has basically the same views as a starter project and the test simply checks that the dev bundle isn't generated. This test is behind a profile and uses local chrome driver. Could you suggest a better test organisation or is it fine?

mshabarov:
The test views can be trimmed to only have a bundle impactful components/view.

vaadin-dev-bundle-test/src/test/java/com/vaadin/devbundle/test/DefaultDevBundleIT.java
```diff
@@ -0,0 +1,56 @@
...
+        waitUntil(driver0 -> $(ButtonElement.class).waitForFirst() != null);
+    }
+
+    @Test
+    public void test() {
+        File baseDir = new File(System.getProperty("user.dir", "."));
+        File devBundle = new File(baseDir, Constants.DEV_BUNDLE_LOCATION);
+
+        // shouldn't create a dev-bundle
+        Assert.assertFalse(devBundle.exists());
```
Artur-, line 47: This could also test that no `node_modules` folder is found

Artur-:
The problem is better fixed by #3778 but the test here should be valid to keep

vaadin-dev-bundle-test/src/test/java/com/vaadin/devbundle/test/DefaultDevBundleIT.java
mshabarov: Done

ZheSun88:
I just added the `dev-bundle` profile to the build.. let us see, how the test goes.

manolo:
We should reuse the test modules that we have already in platform instead of introducing a new module, I'm trying to move this test in this way, but wip right now

manolo:
So closing this since I added dev-mode testing to the platform smoke tests, and included this class https://github.com/vaadin/platform/pull/3788

PR #3787, opened 2023-02-10
test use hub

ZheSun88:


manolo:
not needed after https://github.com/vaadin/platform/pull/3788

PR #3788, opened 2023-02-11
test: add module for testing in dev-mode, and add DefaultDevBundleIT

manolo:
This adds a new module so as we also run smoke tests in dev-mode.
It also adds a test for checking that in dev-mode a fresh project uses the shipped bundle instead of running npm install and compiling frontend

### DIFF
--- a/vaadin-platform-test/pom-dev.xml
+++ b/vaadin-platform-test/pom-dev.xml
@@ -7,7 +7,7 @@
         <artifactId>vaadin-platform-parent</artifactId>
         <version>24.1-SNAPSHOT</version>
     </parent>
-    <artifactId>vaadin-platform-test</artifactId>
+    <artifactId>vaadin-platform-test-dev</artifactId>
     <packaging>war</packaging>
     <properties>
         <maven.compiler.source>17</maven.compiler.source>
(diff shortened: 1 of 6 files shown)


vaadin-platform-test/pom-dev.xml
```diff
@@ -0,0 +1,235 @@
...
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.osgi</groupId>
```
Artur-, line 74: What do you need osgi for?

ZheSun88 (dismissed):

vaadin-platform-test/pom-dev.xml
```diff
@@ -0,0 +1,235 @@
...
+                        <executions>
+                          <execution>
+                            <phase>validate</phase>
+                            <goals>
+                                <goal>run</goal>
+                            </goals>
+                            <configuration>
+                              <target>
+                                <!-- Do not read @Theme annotations from the bundle application class -->
+                                <delete dir="../vaadin-dev-bundle/target/classes/com/vaadin/devbundle" includeemptydirs="true"></delete>
```
Artur-, line 206: This looks super wrong. Why don't you delete the class in the `vaadin-dev-bundle` project instead?

manolo: I tried in the bundle maven install phase, but then maven regenerates the class again in the compile time of the vaadin-platform-test module because it depends on the other.

manolo: Changing the maven phase it seems working, let's see TC

vaadin-platform-test/pom-dev.xml
manolo: Just copied from the default pom.xml, it's needed there because the module is used in servlet container tests.
Although osgi is not supported in 24.0, let's keep that in the other file for now.

Artur- (approved):
much better this way

PR #3800, opened 2023-02-14
chore: rename master to main [skip ci]

ZheSun88:


.github/workflows/sbom.yml
```diff
@@ -1,7 +1,7 @@
 name: SBOM
 on:
   push:
-    branches: ["master", "24.0", "23.3"]
+    branches: ["main", "24.0", "23.3"]
```
manolo, line 4: There are more `master` references in cherryPick which is used in other projects, at some point it could fail, can we figure out somehow the name of the default branch?

ZheSun88: i will do that in another PR.

manolo (approved):

PR #3837, opened 2023-02-24
fix: Use JNA 5.12.1 required by the license checker

Artur-:


mshabarov (approved):

mshabarov:
Would it be convenient if we move the parameter to a more visible place? pom.xml in the root or even versions.json

PR #3874, opened 2023-03-06
chore: align spring dependencies with flow

ZheSun88:
align versions https://github.com/vaadin/flow/blob/main/pom.xml#L87...L90


ZheSun88:
24.0: https://github.com/vaadin/platform/pull/3885

PR #4017, opened 2023-04-05
chore: update jna and slf4j verison to be compatible with license-checker

ZheSun88:


scripts/generator/templates/template-vaadin-bom.xml
```diff
@@ -14,8 +14,8 @@
 
     <properties>
 {{javadeps}}
-        <slf4j.version>1.7.36</slf4j.version>
-        <jna.version>5.12.1</jna.version>
+        <slf4j.version>2.0.7</slf4j.version>
```
Artur-, line 17: This should be 1.x for 23

ZheSun88: reverted that part of change.. seems we have that version in V14 branch too. 

need to check where it comes..

Artur- (approved):

Issue #4046, opened 2023-04-18
Add a new module to platform for building default prod bundle


mshabarov:
Vaadin projects without add-ons (frontend customisations) can skip making it's own production bundle and re-use the one that is built and shipped with the platform and what contains all the Vaadin components. So building the application for production doesn't need Node.js and runs faster.

Acceptance criteria:
- [x] Platform includes a module that produces production bundle in a JAR
- [x] This JAR is added as a dependency to applications automatically, when the application is being built for production

mshabarov:
Closed by https://github.com/vaadin/platform/pull/4049

PR #4049, opened 2023-04-18
feat: Add default production bundle module

mshabarov:
This adds a standalone maven module for default prod bundle. The bundle is generated in a regular way with 'build-frontend' goal. Will be used by Flow in case if an app has no frontend add-ons.

Part-of https://github.com/vaadin/platform/issues/4046

### DIFF
--- a/scripts/generator/templates/template-prod-bundle-pom.xml
+++ b/scripts/generator/templates/template-prod-bundle-pom.xml
@@ -1,0 +1,121 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.vaadin</groupId>
+        <artifactId>vaadin-platform-parent</artifactId>
+        <version>{{platform}}</version>
+    </parent>
+    <artifactId>vaadin-prod-bundle</artifactId>
+    <packaging>jar</packaging>
+    <name>Vaadin Prod Bundle</name>
+    <description>Vaadin Prod Bundle</description>
+    <url>https://vaadin.com</url>
+
+    <distributionManagement>
+        <repository>
+            <id>vaadin-prereleases</id>
+            <url>https://maven.vaadin.com/vaadin-prereleases/</url>
+        </repository>
+    </distributionManagement>
+
+    <properties>
+        {{javadeps}}
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.vaadin</groupId>
+                <artifactId>vaadin-bom</artifactId>
+                <version>${project.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-internal</artifactId>
+            <version>${project.version}</version>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-spreadsheet-flow</artifactId>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <version>2.0.0</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>2.0.0</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>${junit.jupiter.version}</version>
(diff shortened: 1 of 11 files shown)


vaadin-core/pom.xml
```diff
@@ -34,6 +34,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-prod-bundle</artifactId>
```
mshabarov, line 39: Prod bundle is always added, but would be nice to drop it in dev mode and when a custom bundle is generated.

Artur-: Can it be a dependency of the maven and gradle plugins instead of Flow?

mshabarov: I assumed that for using the default prod bundle, no one need to add `vaadin-maven-plugin:build-frontend`, because no frontend build is needed. Thus, if I include it to the plugin module...it wouldn't be included into class-path.

Artur-: You run the production build in the same way also in the future

mshabarov: So you think that `build-frontend` goal is still needed for running the app with default prod bundle?

Artur-: Yes. Even if it is a no-op in some case. That is the place where you would decide what you need to do for the production package

mshabarov: How can I add it exclusively for `build-frontend` goal? `vaadin-maven-plugin` is used in dev mode with `prepare-frontend`, so it will have the default bundle anyway.

Artur-: Why would it have to be exclusively for that? If it is a plugin dependency, you need to copy the files during the production build anyway

vaadin-prod-bundle/src/main/java/com/vaadin/prodbundle/FakeAppConf.java
```diff
@@ -0,0 +1,16 @@
...
+import com.vaadin.flow.component.dependency.JsModule;
+import com.vaadin.flow.component.dependency.NpmPackage;
+import com.vaadin.flow.component.page.AppShellConfigurator;
+import com.vaadin.flow.server.PWA;
+import com.vaadin.flow.theme.Theme;
+
+@Theme("vaadin-prod-bundle")
+@PWA(name = "vaadin-prod-bundle", shortName = "vaadin-prod-bundle")
+@JsModule("@vaadin-component-factory/vcf-nav")
+@NpmPackage(value = "line-awesome", version = "1.3.0")
```
Artur-, line 12:
```suggestion
```

vaadin-core/pom.xml
mshabarov: Done, now it's a dependency of plugins

scripts/generator/templates/template-prod-bundle-pom.xml
```diff
@@ -0,0 +1,147 @@
...
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>com.vaadin</groupId>
+                <artifactId>flow-maven-plugin</artifactId>
+                <version>${flow.version}</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>build-frontend</goal>
```
caalador, line 96:
```suggestion
                            <goal>prepare-frontend</goal>
                            <goal>build-frontend</goal>
```
Build-frontend doesn't like running without prepare-frontend.

scripts/generator/templates/template-prod-bundle-pom.xml
```diff
@@ -0,0 +1,147 @@
...
+                    </goals>
+                    <configuration>
+                      <target>
+                        <!-- Remove application class file, so as IT modules do not read @Theme annotations when
+                             maven configures the classpath for multimodule projects -->
+                          <delete dir="target/classes/com/vaadin/prodbundle"
+                                includeemptydirs="true"/>
+                          <!-- Move bundle files to vaadin-prod-bundle folder -->
+                          <delete dir="target/classes/META-INF"
+                                includeemptydirs="true"/>
```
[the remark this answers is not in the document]
mshabarov: Right, thanks!

scripts/generator/templates/template-prod-bundle-pom.xml
```diff
@@ -0,0 +1,148 @@
...
+                <groupId>com.vaadin</groupId>
+                <artifactId>flow-maven-plugin</artifactId>
+                <version>${flow.version}</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>prepare-frontend</goal>
+                            <goal>build-frontend</goal>
+                        </goals>
+                    </execution>
```
caalador, line 99:
```suggestion
                        </goals>
                        <configuration>
                            <optimizeBundle>false</optimizeBundle>
                        </configuration>
                    </execution>
```
Should the dev bunlde run with optimize false as else it will never be used if you have even one vaadin component in use in the application as with the optimized bundle only found components will be added.

Else it should be fine to use as the starting point.

mshabarov: Production bundle you mean? This template is for production bundle. Build frontend mojo has a default value `true` for it:
```java
    @Parameter(defaultValue = "true")
    private boolean optimizeBundle;
```
Honestly I can't say right now is it needed or not. I'm going to test the checker with default prod bundle soon, so we'll see.

caalador (approved):

scripts/generator/templates/template-prod-bundle-pom.xml
caalador: Yes. Production build always optimizes the bundle to only contain what is found through the endpoints in the application. And with this we do not have anything apart from the Theme annotation making a quite empty bundle.

PR #4061, opened 2023-04-21
chore: remove the snakeyaml exclusion [skip ci]

ZheSun88:


ZheSun88:
we cannot remove the snakeyaml dependency exclusion, because the vulnerability includes all snakeyaml 1.x version.
see sbom run here. https://github.com/vaadin/platform/actions/runs/4762155875

PR #4095, opened 2023-05-02
chore: add metadata to vaadin-quarkus-extension

mcollovati:
The vaadin-quarkus extension artifact is not enough to create a new Vaadin based Quarkus application, as it only depends on flow-server, and it misses jandex index and Vaadin components.
To better integrate with Quarkus tooling, this artifact should be registered within the Quarkus egistry, but it has to contain the Quarkus metadata file that is present in the vaadin-quarkus artifact.


mcollovati:
This change should be ported to every platform version we want to include in the Quarkus Registry.
I suppose 24.0 and 23.3

manolo (approved):

PR #4210, opened 2023-06-01
fix: update plugin versions

caalador:
Update plugin versions used
in vaadin-maven-plugin.

Fixes vaadin/flow#16881


ZheSun88:
should we pick this updates to 24.1 branch too? 

also same for this flow commit https://github.com/vaadin/flow/pull/16911?

mcollovati (approved):

PR #4290, opened 2023-06-28
chore: group platform modules in a profile, remove gradle module activation rule

manolo:
this will simplify the hilla release build by calling the `hilla` profile instead of using `-pl`


ZheSun88 (approved):

ZheSun88:
i do the pick manually

PR #4299, opened 2023-06-29
chore: group platform modules in a profile, remove gradle module activation rule (CP: 24.1)

ZheSun88:
…vation rule


ZheSun88:
24.0 #4301 
23.3 #4305

PR #4564, opened 2023-10-11
remove outdated tomcat exclusions

ZheSun88:
in spring-boot 3.1.4, the integrated version of tomcat-core is 10.1.13, so that the version overwrite is not necessary any more


ZheSun88:
cherry pick to 24.1 branch has been done in #4565

PR #4594, opened 2023-10-17
chore: remove guava/netty dependency overwrite

ZheSun88:


ZheSun88:
The netty version has been updated in selenium 4.14. so currently this removal can only be done in V24.3

PR #4655, opened 2023-11-03
fix: copy dev bundle files from correct place

caalador:
The package feature for devBundle
moves the dev bundle files to a
new folder. Copy from new folder
when generating jar.

Depends on vaadin/flow#17977


ZheSun88:
test failed with 
```
org.opentest4j.AssertionFailedError: The key 'StatisticsGatherer'' should be found in one file ==> expected: <1> but was: <0>
```
also the same failure from main branch with flow 24.3-SNAPSHOT

Artur- (approved):

PR #4776, opened 2023-12-04
chore: update logback-classic to 1.4.14 [skip ci]

ZheSun88:


vaadin-spring-boot-starter/pom.xml
```diff
@@ -76,7 +76,7 @@
         <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
-            <version>1.4.12</version>
+            <version>1.4.14</version>
```
knoobie, line 79: Copy from the closed PR: Got another thing for you :) You should probably also add the other logback-* dependencies, otherwise you get a mismatch which is hard to debug: https://github.com/qos-ch/logback/issues/744#issue-2017777332

vaadin-spring-boot-starter/pom.xml
```diff
@@ -76,7 +76,7 @@
         <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
-            <version>1.4.12</version>
```
knoobie, line 79: I think you have to remove the target/23.x labels because 23.x requires Spring Boot 2.7 where the baseline of logback is 1.2.x (https://github.com/spring-projects/spring-boot/blob/2.7.x/spring-boot-project/spring-boot-dependencies/build.gradle) A fix for the CVE was also added in 1.2.13 (https://github.com/qos-ch/logback/issues/745)

Issue #49, opened 2018-01-16
Add slf4j version to the bom


Artur-:

Artur-:
Uh... that was 5 years ago. Probably to use the correct version? :)

knoobie:
Good argument, I would have said the same thing after 5 years 😅 

I just came across this while trying to upgrade to 2.0.9 (because it finally contains a bom) and was checking where the oldish version came from and there I found this and got my wondering why.. I understand JNA because the whole app can break once it's not the correct version, but slf4j claims to be even binary compatible that I don't see any reason to have it here

PR #4901, opened 2024-01-10
update license-checker to 1.12.4

ZheSun88:


scripts/generator/templates/template-vaadin-bom.xml
```diff
@@ -15,7 +15,7 @@
     <properties>
 {{javadeps}}
         <slf4j.version>2.0.4</slf4j.version>
```
knoobie, line 17: Off Topic: this looks quite old :) 2.0.11 is currently in use in flow / -components 

Additionally, since 2.0.9 you can also use the slf4j-bom instead of managing all dependencies yourself. (Even tho I would ask why Vaadin even tries to enforce a slf4j version to their customer projects :P)

```xml
      <!-- https://central.sonatype.com/artifact/org.slf4j/slf4j-bom -->
      <dependency>
        <groupId>org.slf4j</groupId>
        <artifactId>slf4j-bom</artifactId>
        <version>${slf4j.version}</version>
        <scope>import</scope>
        <type>pom</type>
      </dependency>
```

ZheSun88: I found exactly the same question from https://github.com/vaadin/platform/issues/49 :-P

ZheSun88: I will address this version in a different PR, in case different slf4j versions are used in different branches.

ZheSun88: here is the other one. https://github.com/vaadin/platform/pull/4902

Issue #4936, opened 2024-01-22
Add Vaadin react-component into version.json


mshabarov:
### Describe your motivation

[Vaadin react-components](https://github.com/vaadin/react-components) currently have no split to "core" and "commercial" parts like [Vaadin web-components](https://github.com/vaadin/web-components) do.

That might lead to a problem of leaking commercial react components into an application that only imports `com.vaadin:vaadin-core` artifact.

**Acceptance Criteria:**
- [x] Add react-components into Platform's `version.json` with no split.
- [x] React-components has a split to "core" and "commercial" components possible in the same way as web-components do, e.g. by having the license entry in the `package.json`, see https://github.com/vaadin/react-components/issues/229
- [x] Add `react-components-core` (the name should be discussed, e.g. react-core-components or react-components-all) dependencies into Platform's `version-core.json`.
- [x] Flow then chooses between react or web-components package/dependency in generated `package.json` based on the `reactRouterEnabled` flag in Flow or auto-detection mechanism, see https://github.com/vaadin/flow/issues/18506.

### Additional context

Vaadin/Hilla 24.4.

manolo:
There is a missing task in the description
- [ ] Remove the `vaadin-core-versions.json` files from `hilla` and `hilla-react` artifacts

platform `vaadin-core-versions.json` brings already `hilla-react-components` and flow should select the appropriate entries from here depending on the `reactRouterEnabled` flag or whatever

PR #4952, opened 2024-01-24
draft: add react-component to versions.json

ZheSun88:
related to #4936 

add react-components to vaadin-core-versions.json
add react-components-pro to vaadin-versions.json

depends on vaadin/react-components#229

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -355,7 +355,7 @@
     },
     "react-pro": {
         "react-components-pro": {
-            "jsVersion": "24.4.0-alpha7",
+            "jsVersion": "24.4.0-alpha9",
             "npmName": "@vaadin/react-components-pro"
         }
     },
--- a/versions.json
+++ b/versions.json
@@ -350,6 +350,12 @@
             "npmName": "@vaadin/react-components"
         }
     },
+    "react-pro": {
+        "react-components-pro": {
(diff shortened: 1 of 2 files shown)


versions.json
```diff
@@ -353,6 +353,12 @@
             "npmName": "@vaadin/react-components"
         }
     },
```
manolo, line 355: I feel this should be unified in the same block
```
  react: {
     react-components: {},
     react-components-pro: {}
  }

```

manolo (changes requested):
This PR would be failing until the react is split in two npm packages : https://github.com/vaadin/platform/issues/4936

versions.json
ZheSun88: yeah, it can be. for now, it seems flow is [using the same template](https://github.com/vaadin/flow/pull/18559/files#diff-b1223a5fd26dba93d341bdace64cadce2330f28c958ef3063d5df8a9cfd79243R87) as in this PR. 

the versions.json is only for us to record the versions and most likely to no one is going to hand update the dependency. so should not matter too much.

ZheSun88: now the react components split work has done. 
maybe we should ask for a new release then.

manolo (approved):

Issue #4946, opened 2024-01-23
Include hilla-dev-mode to platform


mshabarov:
### Describe your motivation

Include `hilla-dev-mode` as a dependency to `vaadin-dev`. 
Remove `hilla-dev` from Hilla repository as it wouldn't be needed after that. 

### Additional context

Vaadin/Hilla 24.4.

tltv:
Having `hilla-dev-mode` in `vaadin-dev` ends up some platform's dev mode integration tests to fail. Logs has just warnings that are not there when `hilla-dev-mode`  is not included.

Maybe it's better to leave `hilla-dev-mode` out of the `vaadin-dev` and just add it in hilla repository's `hilla` pom.xml. Or actually just rename `hilla-dev` to `hilla-dev-mode`. ([PR](https://github.com/vaadin/hilla/pull/1992))

If `hilla-dev-mode` has to go to the `vaadin-dev`, then these warnings need to be dealt some other way.

Server side warning:
```
[qtp888293905-94] WARN org.atmosphere.websocket.DefaultWebSocketProcessor - Failed invoking AtmosphereFramework.doCometSupport()
java.lang.NullPointerException: Cannot invoke "org.springframework.context.ApplicationContext.getAutowireCapableBeanFactory()" because the return value of "com.vaadin.hilla.ApplicationContextProvider.getApplicationContext()" is null
        at com.vaadin.hilla.devmode.devtools.DevToolsDatabase.initIfNeeded(DevToolsDatabase.java:48)
        at com.vaadin.hilla.devmode.devtools.DevToolsDatabase.handleConnect(DevToolsDatabase.java:29)
        at com.vaadin.base.devserver.DebugWindowConnection.handleConnect(DebugWindowConnection.java:222)
        at com.vaadin.base.devserver.DebugWindowConnection.onConnect(DebugWindowConnection.java:202)
        at com.vaadin.flow.server.communication.PushHandler.lambda$onConnect$3(PushHandler.java:599)
        at java.base/java.util.Optional.ifPresent(Optional.java:178)
        at com.vaadin.flow.server.communication.PushHandler.lambda$onConnect$4(PushHandler.java:598)
        at com.vaadin.flow.server.communication.PushHandler.callWithServiceAndSession(PushHandler.java:227)
        at com.vaadin.flow.server.communication.PushHandler.onConnect(PushHandler.java:595)
[...]

Issue #4934, opened 2024-01-22
Add Hilla module as a dependency to Platform


mshabarov:
### Describe your motivation

Make `com.vaadin:vaadin` (or `vaadin-core` or `vaadin-spring-boot-starter`, to be discussed and researched) depend on the `com.vaadin.hilla:hilla` artifact, which includes the `endpoint` and `engine-runtime`, so if the `hilla` isn’t desired, one can exclude the whole `hilla` dependency.

### Additional context

Vaadin/Hilla 24.4

knoobie:
Thanks for not adding it to `vaadin-core` :)

mshabarov:
Well, just changed it to `vaadin-core` :)
But indeed I'm not sure 100% what would be the best place for it.
Let me highlight it in the ticket.

knoobie:
5 seconds thought: Hilla requires Spring, therefore it should be in the spring starter as dependency? Vaadin and Vaadin Core does not require Spring

mshabarov:
We cannot really add `hilla` to `vaadin-core` until we split `react-components` into core and all, as requested [https://github.com/vaadin/platform/issues/4936](here).

PR #4953, opened 2024-01-24
draft: add/move hilla dependencies

ZheSun88:
related to #4934 #4946

### DIFF
--- a/vaadin-core-internal/pom.xml
+++ b/vaadin-core-internal/pom.xml
@@ -210,6 +210,11 @@
         </dependency>
 
         <dependency>
+            <groupId>com.vaadin.hilla</groupId>
+            <artifactId>hilla</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>4.13.1</version>
--- a/scripts/generator/templates/template-vaadin-spring-bom.xml
+++ b/scripts/generator/templates/template-vaadin-spring-bom.xml
@@ -26,6 +26,14 @@
     <dependencyManagement>
         <dependencies>
             <dependency>
+                <groupId>com.vaadin.hilla</groupId>
+                <artifactId>hilla-bom</artifactId>
+                <version>${hilla.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+
+            <dependency>
                 <groupId>com.vaadin</groupId>
                 <artifactId>vaadin</artifactId>
                 <version>${project.version}</version>
--- a/vaadin-dev/pom.xml
+++ b/vaadin-dev/pom.xml
@@ -45,6 +45,11 @@
         </dependency>
 
         <dependency>
+            <groupId>com.vaadin.hilla</groupId>
+            <artifactId>hilla-dev-mode</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>com.vaadin</groupId>
             <artifactId>copilot</artifactId>
         </dependency>
--- a/versions.json
+++ b/versions.json
@@ -104,6 +104,9 @@
         "flow": {
             "javaVersion": "24.4.0.alpha3"
         },
+        "hilla": {
+            "javaVersion": "24.4.0.alpha2"
+        },
         "flow-cdi": {
             "javaVersion": "15.0.1"
         },


vaadin-core-internal/pom.xml
```diff
@@ -209,6 +209,11 @@
             <artifactId>vaadin-menu-bar-flow</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>com.vaadin.hilla</groupId>
+            <artifactId>hilla</artifactId>
```
mshabarov, line 214: Would it be better to add `hilla` to `vaadin` until at least we don't have a split for core and all react components?

mshabarov: Also, the idea to add `hilla` to `vaadin-spring-boot-starter` looks not that bad to me now. As Jakarta EE and Quarkus users can just not add `vaadin-spring-boot-starter` to their pom.xml and they won't have Spring and Hilla then.

knoobie: This would also make it pretty easy to exclude hilla by not using `vaadin-spring-boot-starter` because it is only a wrapper for `vaadin-spring`, `vaadin-core` and `hilla`. So the flow developer only has to add `vaadin-spring` and `vaadin-core` if they don't wanna deal with exclusions.

mshabarov, line 214: Also I think we have to make changes in `hilla` artifact, so here are some possible actions:

1. `pom.xml` can be retained as is in `hilla`.
2. `@NpmPackage` annotations should be removed from `hilla`.
3. `hilla-react` can be just removed from Hilla as it has the same pom.xml as `hilla`.
4. `hilla` then is added to Platform.
5. Flow adds these three react packages (Auto Crud etc.) if `reactRouterEnabled=true` otherwise it adds one lit form package (in `NodeUpdater` class).

knoobie: > Flow adds these three react packages (Auto Crud etc.) if reactRouterEnabled=true otherwise it adds one lit form package (in NodeUpdater class).

I would vote against this; or add "if hilla is on the class path" - otherwise flow only apps gets react packages thanks to reactRouterEnabled=true as default

tltv:
Build fails due to `hilla-bom` is not defining `hilla-dev-mode`.

vaadin-core-internal/pom.xml
tltv: This would also give easy option to divide it in two platform modules like `vaadin-spring-boot-starter` and `vaadin-hilla-spring-boot-starter`.

mshabarov: The idea is to use one dependency `vaadin-spring-boot-starter` and avoid "hilla" artifacts, so having a second one with the hilla in the middle is approx. the same as have old `hilla-spring-boot-starter`.

mshabarov: > or add "if hilla is on the class path"

Yes, agree, that's a good point.

I proposed to remove npm package annotations and add packages imperatively in NodeUpdater, so that to relief developers from excluding artifacts by themselves.

mshabarov: Updated the wording for 5.

manolo (approved):

scripts/generator/templates/template-vaadin-bom.xml
```diff
@@ -115,6 +115,13 @@
                 <artifactId>jna-platform</artifactId>
                 <version>${jna.version}</version>
             </dependency>
+
+            <!-- workaround,should be removed after this module has been added to hilla-bom -->
+            <dependency>
+                <groupId>com.vaadin.hilla</groupId>
+                <artifactId>hilla-dev-mode</artifactId>
```
mshabarov, line 122: Maybe I missed something important, but how with this structure one can exclude all development related dependencies in hybrid app ? I expected only one step as previously - excluding `vaadin-dev` in pom.xml.
See my comment [here](https://github.com/vaadin/platform/issues/4946#issuecomment-1914431757).

ZheSun88: this is only a bom, which doesn't bring in any dependency directly, as it is only under dependencyManagement. 

also for this case, it is a workaround as `hilla-dev-mode` is not in `hilla-bom` yet

mshabarov: Yep, agree, maybe this isn't a better place to ask this.

mshabarov: However, not clear to me how Hilla dev things should be excluded in prod mode, and should they even be excluded?

PR #4966, opened 2024-01-30
chore: vaadin-spring-boot-starter should bring the hilla-react by default instead of hilla

manolo:
User needs to exclude hilla-react and add hilla if they want to use hilla-lit instead of hilla-react

### DIFF
--- a/vaadin-spring-boot-starter/pom.xml
+++ b/vaadin-spring-boot-starter/pom.xml
@@ -55,7 +55,7 @@
         </dependency>
         <dependency>
             <groupId>com.vaadin.hilla</groupId>
-            <artifactId>hilla</artifactId>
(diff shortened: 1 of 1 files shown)


ZheSun88:
the change in this PR is different than the description in this issue ticket, https://github.com/vaadin/platform/issues/4934
so everything okay about this change on flow side @mshabarov 

```
Make com.vaadin:vaadin (or vaadin-core or vaadin-spring-boot-starter, to be discussed and researched) depend on the com.vaadin.hilla:hilla artifact
```

mshabarov:
Yes, we discussed this at some meeting. Let's go with the change in this PR, let me update the ticket.

ZheSun88 (approved):

PR #4968, opened 2024-01-30
chore: add hilla-dev-mode in vaadin-dev

tltv:
Fixes: #4946

Note! Need to merge first Hilla's https://github.com/vaadin/hilla/pull/2001.

### DIFF
--- a/scripts/generator/templates/template-prod-bundle-pom.xml
+++ b/scripts/generator/templates/template-prod-bundle-pom.xml
@@ -33,6 +33,11 @@
             <optional>true</optional>
         </dependency>
         <dependency>
+            <groupId>com.vaadin.hilla</groupId>
+            <artifactId>hilla-dev-mode</artifactId>
+            <version>${hilla.version}</version>
+        </dependency>
+        <dependency>
             <groupId>com.vaadin</groupId>
             <artifactId>vaadin-spreadsheet-flow</artifactId>
             <version>{{platform}}</version>
--- a/scripts/generator/templates/template-prod-bundle-pom.xml
+++ b/scripts/generator/templates/template-prod-bundle-pom.xml
@@ -33,11 +33,6 @@
             <optional>true</optional>
         </dependency>
         <dependency>
-            <groupId>com.vaadin.hilla</groupId>
-            <artifactId>hilla-dev-mode</artifactId>
-            <version>${hilla.version}</version>
-        </dependency>
-        <dependency>
             <groupId>com.vaadin</groupId>
             <artifactId>vaadin-spreadsheet-flow</artifactId>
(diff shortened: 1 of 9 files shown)


ZheSun88 (approved):

vaadin-dev/pom.xml
```diff
@@ -44,6 +44,11 @@
             <artifactId>vaadin-dev-bundle</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>com.vaadin.hilla</groupId>
+            <artifactId>hilla-dev-mode</artifactId>
+        </dependency>
+
```
manolo, line 51: why is this needed here instead of on the spring starter ?

manolo (changes requested):

vaadin-dev/pom.xml
tltv: Discussed in slack. Conclusion: remove it from spring starter.

vaadin-platform-test/src/test/java/com/vaadin/platform/test/ChromeComponentsIT.java
```diff
@@ -83,8 +83,13 @@ protected String getTestPath() {
     }
 
     @Before
-    public void load(){
-        $(NotificationElement.class).waitForFirst();
+    public void load() {
+        if (getDriver().getPageSource().contains("<div class=\"message\">Building front-end development bundle</div>")) {
+            // Wait for the development bundle to be loaded
+            $(NotificationElement.class).waitForFirst(180);
```
manolo, line 89: I feel that ff the problem is that we need for the bundle to be compiled, probably this fix should be in the testbench instead

scripts/generator/templates/template-prod-bundle-pom.xml
```diff
@@ -32,6 +32,11 @@
             <version>${project.version}</version>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>com.vaadin.hilla</groupId>
+            <artifactId>hilla-dev-mode</artifactId>
```
mshabarov, line 37: This shouldn't go into prod pre-compiled bundle, as it brings HotSwapAgent plugin and vaadin-dev-server

mshabarov (changes requested):

scripts/generator/templates/template-prod-bundle-pom.xml
tltv: I agree. And hilla-dev-mode should be excluded from production builds in general. This is not automatic however but build configs can exclude whole vaadin-dev from the production build.

manolo (approved):

PR #4978, opened 2024-02-01
fix vaadin gradle plugin

taefi:
This adds changes for:

- extracting sources from hilla-gradle-plugin sources artifact to a temp dir
- copying sources from temp dir to src/main/kotlin/com/vaadin/gradle
- renaming HillaPlugin to VaadinPlugin
- updating package declarations in sources according to the new package

### DIFF
--- a/scripts/generator/templates/template-vaadin-gradle-plugin-pom.xml
+++ b/scripts/generator/templates/template-vaadin-gradle-plugin-pom.xml
@@ -40,6 +40,11 @@
         </dependency>
         <dependency>
             <groupId>com.vaadin</groupId>
+            <artifactId>flow-gradle-plugin</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
             <artifactId>vaadin-prod-bundle</artifactId>
             <version>${project.version}</version>
         </dependency>
@@ -49,6 +54,12 @@
             <groupId>com.vaadin</groupId>
             <artifactId>vaadin-dev-bundle</artifactId>
             <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.groovy</groupId>
+            <artifactId>groovy-all</artifactId>
+            <version>4.0.18</version>
+            <type>pom</type>
         </dependency>
     </dependencies>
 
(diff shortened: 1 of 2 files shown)


ZheSun88:
ah, the flow-gradle-plugin should not use the project version in this case, as flow has its own version.. 
i will modify the PR a bit to let it work.

taefi:
> ah, the flow-gradle-plugin should not use the project version in this case, as flow has its own version.. i will modify the PR a bit to let it work.

Yeah, that I was about to ask you to fix, thanks.
This makes the final vaadin-gradle-plugin to look much like what it was before when it comes to the namings, packaging, and the way people can use it in their applications.

ZheSun88:
now seems a different error. 
```
 sourceFile /opt/agent/work/9b1a81aa43ab65b4/vaadin-gradle-plugin/src/main/kotlin/com/vaadin/gradle/HillaPlugin.kt does not exist
```

ZheSun88:
I wonder if the groovy plugin used here has some conflict with the one here. https://github.com/vaadin/platform/blob/f1456ca359831949a3b0f1d38294cf6dc8fa3a91/vaadin/pom.xml#L65...L87

scripts/generator/templates/template-vaadin-gradle-plugin-pom.xml
```diff
@@ -113,6 +118,86 @@
...
+                            <scripts>
+                                <script>
+                                    <![CDATA[
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.nio.file.Path
+import java.util.stream.Collectors
+
+// Define the directory to search for Kotlin files
+String directoryPath = 'src/main/kotlin/com/vaadin/gradle/plugin'
```
ZheSun88, line 169: it seems this path is the reason that failed the build.
if i update it as `vaadin-gradle-plugin/src/main/kotlin/com/vaadin/gradle/plugin`.. the build works..

if i dont have the `vaadin-gradle-plugin/` , running from the `gradle-plugin` module works.

seems here should use a relative path instead

ZheSun88: validated using this PR https://github.com/vaadin/platform/pull/4987.
it works..

ZheSun88 (approved):

PR #4983, opened 2024-02-02
chore: Copilot version update to 24.3.7

czp13:
Dependency update only


Artur-:
Let's not, it is broken

PR #4986, opened 2024-02-02
fix: npm install error Unable to resolve reference $@types/react-dom

platosha:


ZheSun88:
i think we can close this PR now. 
reopen it if needs

PR #5001, opened 2024-02-05
fix: include React components with npm install in -sbom

platosha:
closes #4986

Alternative cleaner solution to fix npm install error.


platosha:
OK this did not work...

PR #5055, opened 2024-02-15
feat: add flow-react module

caalador:
add flow-react into core-internals

Requires vaadin/flow#18710

### DIFF
--- a/vaadin-core-internal/pom.xml
+++ b/vaadin-core-internal/pom.xml
@@ -45,6 +45,10 @@
         </dependency>
         <dependency>
             <groupId>com.vaadin</groupId>
+            <artifactId>flow-react</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
             <artifactId>flow-push</artifactId>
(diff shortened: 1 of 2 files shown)


tltv (approved):

vaadin-core-internal/pom.xml
```diff
@@ -43,6 +43,10 @@
             <groupId>com.vaadin</groupId>
             <artifactId>flow-lit-template</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-react</artifactId>
```
ZheSun88, line 48: should add to the bom for the version management.

ZheSun88: pushed the commit

caalador: So many places... I sort of expected it to come from the flow-bom

ZheSun88: one thing we can start to use in platform is import flow-bom. same as `hilla-bom`

PR #5076, opened 2024-02-20
chore: add mode property to versions.json

tltv:
Adds "mode": "lit" to core and pro web-components. Flow will need this information to update package.json dynamically. All dependencies in lit mode are already included in react-components dependency.

Related-to: vaadin/flow#18684

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -8,98 +8,118 @@
     "core": {
         "a11y-base": {
             "jsVersion": "24.4.0-alpha13",
-            "npmName": "@vaadin/a11y-base"
+            "npmName": "@vaadin/a11y-base",
+            "mode": "lit"
         },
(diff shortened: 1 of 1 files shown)


versions.json
```diff
@@ -350,47 +406,55 @@
     "react": {
         "react-components": {
             "jsVersion": "24.4.0-alpha13",
-            "npmName": "@vaadin/react-components"
+            "npmName": "@vaadin/react-components",
```
mshabarov, line 409: The overall "react-components" json can be moved to "vaadin" I believe to be consistent with the overall structure. But this might be a breaking change for platform, so let's make it in another PR at some point. Not a blocker at all.

mshabarov (approved):

PR #5093, opened 2024-02-26
chore: add exclusion list to react-components

manolo:
It would allow flow to know the list of frontend dependencies that should not be added to package.json file when react-components is used


versions.json
```diff
@@ -407,7 +407,72 @@
...
+                "@vaadin/context-menu",
+                "@vaadin/cookie-consent",
+                "@vaadin/crud",
+                "@vaadin/custom-field",
+                "@vaadin/date-picker",
+                "@vaadin/date-time-picker",
+                "@vaadin/details",
+                "@vaadin/dialog",
+                "@vaadin/email-field",
+                "@va4.din/field-highlighter",
```
knoobie, line 434:
```suggestion
                "@vaadin/field-highlighter",
```

Artur-:
Would it not be better to read the actual file and write the content during the build?

manolo:
> Would it not be better to read the actual file and write the content during the build?

Well, this seems simpler than maintaining a function in a script doing this, it might be that we can have certain dependencies in react-components that is not necessary to copy in this block.
Let's keep in this way right now so as flow can use the resulting file, and if maintaining the list here begins a problem we can modify the vaadin-version generator script

mshabarov (approved):

PR #5104, opened 2024-02-27
fix: Remove non-components deps from exclusion list

mshabarov:
These mixins, when excluded, are installed into `node_modules`, but the overall theming is broken and the lumo styles are not applied to UI:

Screenshot 2024-02-27 at 12 11 01


tltv (approved):

mshabarov:
Probably it was a glitch on my local, as I cannot reproduce anymore.

vursen:
> Probably it was a glitch on my local, as I cannot reproduce anymore.

I can reproduce it in my flow starter app.

https://github.com/vaadin/platform/assets/5039436/f180158a-5375-4ae6-8f38-3203aef2962d

manolo (approved):

PR #5106, opened 2024-02-27
test: re-enable tests checking usage of bundle

manolo:
These tests were temporary disabled during the unification of hilla in platform, and needs to be re-enabled before beta1


manolo:
This needs a fix in flow for this issue https://github.com/vaadin/flow/issues/18817

Issue #5230, opened 2024-04-12
Optimize Platform artifacts fo most commonly used cases and make it more flexible to configure


mshabarov:
### Describe your motivation

Vaadin 24.4 unifies Flow and Hilla frameworks that, among other things, means extra dependencies tree in the `com.vaadin:vaadin` and `com.vaadin:vaadin-core` artifacts, which aren't always necessary and desired for projects. With the current structure this may lead to Maven exclusions in the projects, that aren't optimal and clean enough.

This feature request proposes a list of new artifacts in Vaadin Platform that should simplify picking the minimal set of Vaadin products needed for a project and allow developers to opt-in the necessary dependencies, instead of opt-out with Maven exclusions.

For the most used projects set-ups Vaadin Platform provides the "starters" dependencies, that includes all needed Vaadin dependencies for a target framework/container. From the other side, Vaadin Platform provides more granular artifacts that one can combine to build a desired configuration. The number of various Vaadin products and other frameworks, that Vaadin has integration with, may lead to the long list of different combination, but we'll focus on the key ones and assumes that for other combinations one would need to add a bit more granular artifacts, say one up to three.

Backwards compatibility is preserved, i.e. we have to preserve the old artifacts that developers are used to have, but deprecate them and recommend to use the new ones.

**Acceptance Criteria**

- [ ] ( 1 ) `com.vaadin:vaadin` and `com.vaadin:vaadin-core` are be deprecated (in documentation and log a warning message in runtime), but not removed.
 - [ ] ( 2 ) Changes in 24.4 regarding extra dependencies in `com.vaadin:vaadin` and `com.vaadin:vaadin-core` (e.g. adding Hilla into `vaadin-core`) should be reverted back, so these artifacts would be backwards compatible and not give unexpected errors or behaviour to users who upgrade to 24.4.
- [ ] ( 3 )Vaadin groups core and pro Flow components into separate dependencies, so they can be added to projects and Platform dependencies separately: `flow-components` (free) and `flow-components-pro` (commercial). These names are inline with the React components: `react-components` (free) and `react-components-pro` (commercial).
[...]

PR #5264, opened 2024-04-24
feat: Create vaadin-dev-flow and vaadin-dev-hilla

### LINKED ISSUE #5260: Move Hilla from vaadin to vaadin-spring-boot-starter artifact in Vaadin 24.4
### Describe your motivation

Preserve backwards compatibility for the `vaadin` and `vaadin-core` dependencies for, e.g. Quarkus users by not including Hilla in them.

### Describe the solution you'd like

Add Hilla as a dependency of `vaadin-spring-boot-starter` artifact of Vaadin Platform. Spring Boot users who want to optimize their deployment size to not include Hilla can exclude it from that dependency.

### Describe alternatives you've considered

_No response_

### Additional context

Planned for Vaadin 24.4.0.beta2 or any further beta release.

caalador:
vaadin-dev no has vaadin-dev-flow and vaadin-dev-hilla.
vaadin-core depends on vaadin-dev-flow
to not get hilla if not wanted.

This will make vaadin and vaadin-core backwards
compatible with older version where hilla was not a dependency.

Implements part of #5230

Fixes #5260

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,8 @@
                 <module>vaadin-bom</module>
                 <module>vaadin-core-internal</module>
                 <module>vaadin-core</module>
+                <module>vaadin-dev-flow</module>
+                <module>vaadin-dev-hilla</module>
                 <module>vaadin-dev</module>
                 <module>vaadin-maven-plugin</module>
                 <module>vaadin-testbench</module>
--- a/scripts/generator/templates/template-vaadin-spring-bom.xml
+++ b/scripts/generator/templates/template-vaadin-spring-bom.xml
@@ -45,6 +45,16 @@
             </dependency>
             <dependency>
                 <groupId>com.vaadin</groupId>
+                <artifactId>vaadin-dev-flow</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.vaadin</groupId>
+                <artifactId>vaadin-dev-hilla</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.vaadin</groupId>
                 <artifactId>vaadin-dev</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/scripts/generator/templates/template-dev-bundle-pom.xml
+++ b/scripts/generator/templates/template-dev-bundle-pom.xml
@@ -33,12 +33,6 @@
         </dependency>
         <dependency>
             <groupId>com.vaadin</groupId>
-            <artifactId>hilla-dev</artifactId>
-            <version>${hilla.version}</version>
-            <optional>true</optional>
-        </dependency>
-        <dependency>
-            <groupId>com.vaadin</groupId>
             <artifactId>vaadin-spreadsheet-flow</artifactId>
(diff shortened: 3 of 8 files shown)


pom.xml
```diff
@@ -66,6 +66,8 @@
                 <module>vaadin-bom</module>
                 <module>vaadin-core-internal</module>
                 <module>vaadin-core</module>
+                <module>vaadin-dev-flow</module>
+                <module>vaadin-dev-hilla</module>
```
mshabarov, line 70: Not really a comment against this design, but just a concern how backwards compatible will this approach be: in existing projects the artifact `vaadin-dev` is excluded from prod builds. Now, as only `vaadin-dev-flow` is included into `vaadin` and `vaadin-core`, the old approach with excluding `vaadin-dev` will have no effect (I suppose Maven work in this way).

So we either have to document this clearly, or let vaadin-dev be semantically `vaadin-flow-dev`, which I don't really like.

caalador: We should have a clear mention in the upgrade notes that it is named vaadin-dev-flow in 24.4

Also as the vaadin(-core) packages are getting "deprecated" it should be noted how the new package structure is going to be used in the update notes also.

mshabarov: As we discussed offline, we can set the `hilla-dev` dependency as `optional` in `vaadin-dev`, which basically would relief us from the need of splitting `vaadin-dev` into two parts.
Then `vaadin` and `vaadin-core` wound't get Hilla dev things and in `vaadin-spring-boot-starter` we can explicitly add `hilla-dev` in the list of dependencies. 
Finally `vaadin-dev` exclusion should work in a backwards compatible way after that.

mshabarov: > Also as the vaadin(-core) packages are getting "deprecated" it should be noted how the new package structure is going to be used in the update notes also.

Not sure that in 24.4 we are going to deprecate them, as we don't provide an alternative in 24.4.
I think once we introduce the new starter artifacts, then we should deprecate.

scripts/generator/templates/template-vaadin-spring-bom.xml
```diff
@@ -43,6 +43,16 @@
...
+            <dependency>
+                <groupId>com.vaadin</groupId>
+                <artifactId>vaadin-dev-flow</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.vaadin</groupId>
+                <artifactId>vaadin-dev-hilla</artifactId>
+                <version>${project.version}</version>
+            </dependency>
```
mshabarov, line 55: A new dilemma: `vaadin-dev-hilla` is inside deps management here and it has to be added to `<dependencies>` in `vaadin-spring-boot-starter` to become a part of it. The as is structure doesn't add `vaadin-dev-hilla` into Vaadin Spring projects.

But if we add it, we no more could exclude it in prod mode by excluding `vaadin-dev`.

mshabarov:
An alternative solution is proposed here https://github.com/vaadin/hilla/pull/2383. This PR may become a refactoring PR that doesn't move `hilla-dev`, but just splits `vaadin-dev` into two artifacts.

PR #5833, opened 2024-05-09
chore: align hilla and flow versions with main 24.5-SNAPSHOT

manolo:
It will have a deployed version when we have the first alpha for 24.5


ZheSun88 (approved):

ZheSun88:
will ask for the copilot 24.5-SNAPSHOT

PR #5834, opened 2024-05-09
chore: add package-lock.json for hybrid apps to dev-bundle

mcollovati:
A 'hybrid-package-lock.json', containing also Hilla dependencies, is added to the vaadin-dev-bundle artifact, so that Flow can choose the appropriate lock file based on the project configuration.

Part of vaadin/flow#19298


mcollovati:
The main problem here is that the hybrid module triggers Hilla tasks, that in turn are trying to run Maven to execute the `vaadin:configure` goal, that fails because the plugin cannot be found.

```
[ERROR] No plugin found for prefix 'vaadin' in the current project and in the plugin groups [org.sonatype.plugins, org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/home/user/.m2/repository), central (https://repo.maven.apache.org/maven2), vaadin-prereleases (https://maven.vaadin.com/vaadin-prereleases)] -> [Help 1]
```

Causes:

* the `vaadin-maven-plugin` is not configured int the module, and it cannot. Otherwise, there will be a cyclic reference
* the current Vaadin version cannot be detected by Hilla because `vaadin-core` is not a dependency of the module (Hilla reads the version from the artifact `pom.propeties` file and then runs the goal with full artifact coordinates), and it cannot. Otherwise, there will be a cyclic reference

mcollovati:
So, to recap:

`vaadin-hybrid-dev-bundle` should be built before `vaadin-dev-bundle` to be able to grab the `package-lock.json` with Hilla dependencies. However:

- `vaadin-hybrid-dev-bundle` runs Hilla that requires `vaadin-maven-plugin`
- `vaadin-maven-plugin` requires `vaadin-dev-bundle`
- `vaadin-dev-bundle` requires `vaadin-hybrid-dev-bundle`

mcollovati:
The trick is to exclude `hilla-engine-runtime`, so Hilla generators are not executed, but the `package-lock.json` with Hilla NPM dependencies is correctly created.

manolo (approved):

PR #6519, opened 2024-06-18
feat: Property to disable hilla task registration

caalador:
Fixes vaadin/flow#19573

### DIFF
--- a/scripts/generator/templates/template-vaadin-gradle-plugin-pom.xml
+++ b/scripts/generator/templates/template-vaadin-gradle-plugin-pom.xml
@@ -174,6 +174,9 @@ String oldPackage = 'package com.vaadin.hilla.gradle.plugin'
 String newPackage = 'package com.vaadin.gradle.plugin'
 String oldPluginClassDef = 'public class HillaPlugin : Plugin<Project> {'
 String newPluginClassDef = 'public class VaadinPlugin : Plugin<Project> {'
+String importDef = 'import com.vaadin.gradle.VaadinPlugin'
+String projectDef = 'override fun apply(project: Project) {'
+String projectTaskDef = 'project.tasks.apply {'
 
 // Use NIO to walk through the directory
 Files.walk(Paths.get(directoryPath))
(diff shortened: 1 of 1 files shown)


caalador:
After this patch to get rid of hilla and spring add to `gradle.properties`  `hilla.active=false` (op on the commandline `gradlew [tasks]  -Philla.active=false`)

Exclude hilla-engine-core in build.gradle buildscript
```
buildscript {
    configurations.classpath {
        exclude group: 'com.vaadin', module: 'hilla-engine-core'
    }
}
```
Exclude hilla-dev from vaadin(-core)  import
```
    implementation ("com.vaadin:vaadin-core") {
        exclude group: 'com.vaadin', module: 'hilla-dev'
    }
```

caalador:
Testing the plugin as a local build run 
- `scripts/generateBoms.sh`
- `mvn package` in the vaadin-gradle module
- `./gradlew clean build publishToMavenLocal` in vaadin-gradle module.

Add to the test project `build.gradle` on top the buildscript:
```
buildscript {
    repositories {
        mavenLocal()
    }
    dependencies {
        classpath group: 'com.vaadin',
                name: 'vaadin-gradle-plugin',
                version: '24.4-SNAPSHOT'
    }
    configurations.classpath {
        exclude group: 'com.vaadin', module: 'hilla-engine-core'
    }
}
```

mcollovati (commented):
Would it be possible to autodetect Hilla usage as we do in the Flow maven plugin?

```
    public static boolean isHillaUsed(MavenProject mavenProject,
            File frontendDirectory) {
        return isHillaAvailable(mavenProject)
                && FrontendUtils.isHillaViewsUsed(frontendDirectory);
    }
    public static boolean isHillaAvailable(MavenProject mavenProject) {
        List<String> classpathElements = FlowModeAbstractMojo
                .getClasspathElements(mavenProject);
        return BuildFrontendUtil.getClassFinder(classpathElements).getResource(
                "com/vaadin/hilla/EndpointController.class") != null;
    }
```

The `ClassFinder` and `frontendDirectory` could probably be taken from Flow's `GradlePluginAdapter`.

I wonder if we can add similar methods in the Flow gradle plugin (e.g. `VaadinUtils` or `GradlePluginAdapter` itself) and use them in the groovy script in the POM file.

caalador:
I tried that, but I couldn't get the plugin to collect the classpath data without gradle throwing
```
* What went wrong:
A problem occurred evaluating root project 'gradle-validation'.
> Cannot change dependencies of dependency configuration ':implementation' after it has been included in dependency resolution.
```
I guess all the plugin things should be run inside a `project.afterEvaluate {` block.
Also there is a lot that is private making the needed code for collecting the classpath quite long, so getting to the data with a replace string doesn't sound tempting.

mcollovati:
I see. Thanks for clarification

mcollovati (approved):

mshabarov (approved):
Tested with JDK 21, `./gradlew tasks` works for me with the described exclusion and `hilla.active=false`.

ZheSun88:
great, if this works as expected and fix the reported issue, let us get it in.. 

but i think we probably should document this usage somewhere, where do you think? release note or docs?

mshabarov:
Should be documented in online docs, see this PR https://github.com/vaadin/docs/pull/3499

mshabarov:
And also Flow-Gradle starters will be updated to exclude Hilla.

ZheSun88:
great.. then let us merge this?

PR #6679, opened 2024-08-19
chore: add control center to V24.5

ZheSun88:


ZheSun88:
```
pkg:maven/com.squareup.okhttp3/logging-interceptor@3.12.12
```
is from control center, as it depends on an older version of springboot. PR to upgrade sb has been made https://github.com/vaadin/control-center/pull/440

PR #6803, opened 2024-09-18
[test PR] test copilot 24.5-SNAPSHOT

ZheSun88:


ZheSun88:
confirmed the issue is with the orange circle from copilot

PR #6865, opened 2024-10-11
test: Add react mixed test module

### LINKED ISSUE #4972: New React hybrid application test module
### Describe your motivation

Add a new test module into platform and smoke tests for React hybrid application.

Existing Lit hybrid app can be used for a reference and also https://www.github.com/vaadin/flow-hilla-hybrid-example.

### Additional context

Vaadin 24.4.

caalador:
Add a mixed React Hilla Flow
test module to test interoperability
of the 2 features

Fixes #4972


vaadin-platform-react-hybrid-test/src/main/frontend/views/hilla/components.tsx
```diff
@@ -0,0 +1,315 @@
...
+    VirtualList,
+    VirtualListItemModel
+} from "@vaadin/react-components";
+import type { ViewConfig } from "@vaadin/hilla-file-router/types.js";
+import {
+    Board,
+    BoardRow,
+    Chart,
+    ChartSeries,
+    CookieConsent,
```
ZheSun88, line 63: there is a new component called, dash board in V24.6.. 
should we add it here?

caalador: We could, This just does the same test as lit hybrid that there are no exceptions thrown....

vaadin-platform-react-hybrid-test/src/main/java/com/vaadin/platform/react/test/views/FlowMainView.java
```diff
@@ -0,0 +1,27 @@
+package com.vaadin.platform.react.test.views;
+
+import com.vaadin.flow.component.button.Button;
+import com.vaadin.flow.component.html.Span;
+import com.vaadin.flow.component.orderedlayout.VerticalLayout;
+import com.vaadin.flow.router.Route;
+import com.vaadin.flow.router.RouterLink;
+
+@Route("flow")
+public class FlowMainView extends VerticalLayout {
```
mshabarov, line 10: I don't know why this doesn't happening in the validation of this PR, but locally I get 
```
Caused by: com.vaadin.flow.server.InvalidRouteConfigurationException: Invalid route configuration. The following Hilla route(s) conflict with configured Flow routes: flow
```
because this view has apparently the same mapping as `views/flow/@index.tsx`.

mshabarov: Even if I remove `views/flow/@index.tsx`, it still throws, because apparently it sees an entry with "flow" path in the `file-routes.json`.

mshabarov: I fixed it by using "flow-flow" mapping for Flow view.

mshabarov (commented):
I couldn't fine the following test cases:
1. Navigation lifecycle events handling on server: that event is triggered with expected number of times
2. Server-side API for browser navigation: History::back/forward, History::pushState/replaceState

Do you think this isn't needed?

vaadin-platform-react-hybrid-test/src/main/java/com/vaadin/platform/react/test/views/FlowMainView.java
caalador: There was the check for duplicate routes PR just a while ago and this one is older than that.
But just removing the `@index.tsx` from the flow folder should be enough. Probably the mapping check needs some change on this.

vaadin-platform-react-hybrid-test/src/main/frontend/utils/css-utils.ts
```diff
@@ -0,0 +1,23 @@
+import { DomModule } from "@polymer/polymer/lib/elements/dom-module";
+import { stylesFromTemplate } from "@polymer/polymer/lib/utils/style-gather";
+import { CSSResult, unsafeCSS } from "lit";
+
+/**
+ * Utility function for importing style modules. This is a temporary
+ * solution until there is a standard solution available
+ * @see https://github.com/vaadin/vaadin-themable-mixin/issues/73
```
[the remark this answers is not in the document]
caalador: Probably can be removed. I just started with a full copy of the original hybrid module.

vaadin-platform-react-hybrid-test/src/main/java/com/vaadin/platform/react/test/views/FlowMainView.java
mshabarov: Just removing it doesn't help. Perhaps we have a bug in conflicts detection in Flow.

mshabarov: Ok, validation passes.

mshabarov (approved):

caalador:
I'm not certain it is run as it is behind the profile. Same as the other hybrid which I don't know where it is executed.
```
        <profile>
            <id>react-hybrid</id>
            <modules>
                <module>vaadin-platform-react-hybrid-test</module>
            </modules>
        </profile>
```

@ZheSun88  any idea on the execution?

mshabarov:
At least yesterday I saw that these IT tests were running according to the log.

pom.xml
```diff
@@ -76,6 +76,7 @@
                 <module>vaadin-dev-bundle</module>
                 <module>vaadin-prod-bundle/pom-unoptimized.xml</module>
                 <module>vaadin-prod-bundle</module>
+                <module>vaadin-platform-react-hybrid-test</module>
```
mshabarov, line 79: The test module is behind a profile and added in the standard list of modules, maybe we should excluded it from here and leave it only behind a profile?

caalador: Yes.

mshabarov (approved):

mshabarov:
@ZheSun88 could you please add this profile to CI when appropriate?

PR #6899, opened 2024-10-17
test: Flow-Hilla with security module

caalador:
Add test module for Flow-Hilla
mixed application with security
configuration.

part of https://github.com/vaadin/platform/issues/4972


pom.xml
```diff
@@ -95,6 +95,12 @@
                 <module>vaadin-platform-hybrid-test</module>
             </modules>
         </profile>
+        <profile>
+            <id>hilla-hybrid</id>
```
mshabarov, line 99: "hybrod-security" would be a better name probably.

mshabarov (approved):

pom.xml
caalador: done

mshabarov:
@ZheSun88 could you please add this profile into CI when appropriate?

PR #6961, opened 2024-11-04
chore: add vaadin-core-components module

tltv:
Adds vaadin-core-components module that contains only core component dependencies. Makes it easier to exclude all core components on demand from vaadin-core module.

RelatedTo: vaadin/flow#19948


caalador (commented):
Seems ok. Zhe had the suggestion of creating the core-components module in flow-components repository instead of platform.

It would make sense in the way that when adding new components it would be easier to update in the same place.

tltv:
> Zhe had the suggestion of creating the core-components module in flow-components repository instead of platform.
> 
> It would make sense in the way that when adding new components it would be easier to update in the same place.

It would make sense if all core components were in one repository, but `flow-components` don't have all core components. At least `collaboration-engine` is one of those. It used to be pro component, but was moved to core later. It's more like a composite than web component, but it depends on other core components like avatar.

scripts/generator/templates/template-vaadin-spring-bom.xml
```diff
@@ -78,6 +78,11 @@
                 <artifactId>vaadin-prod-bundle</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
```
Artur-, line 81: This already has a core dependency. What does this add?

tltv: This is needed here to tell bom the version of the vaadin-core-components.

caalador (approved):

PR #7056, opened 2024-11-29
feat: Add test for mixed flowLayout and deep tree

caalador:
Add tests for testing use of
flowLayout as both true and false
in same layout path.
Add test for hilla view with a deep
path that should get a Flow layout.

### DIFF
--- a/vaadin-platform-react-hybrid-test/src/test/java/com/vaadin/platform/react/test/FlowMainLayoutIT.java
+++ b/vaadin-platform-react-hybrid-test/src/test/java/com/vaadin/platform/react/test/FlowMainLayoutIT.java
@@ -8,6 +8,7 @@ import org.openqa.selenium.support.ui.ExpectedConditions;
 
 import com.vaadin.flow.component.button.testbench.ButtonElement;
 import com.vaadin.flow.component.html.testbench.AnchorElement;
+import com.vaadin.flow.component.sidenav.testbench.SideNavItemElement;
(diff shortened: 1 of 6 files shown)


vaadin-platform-react-hybrid-test/src/test/java/com/vaadin/platform/react/test/FlowMainLayoutIT.java
```diff
@@ -16,13 +17,49 @@ public void hillaViewInFlowLayout() {
...
-        // navigate away from Flow view
+        // navigate away from Hilla view
+        getMenuElement("Flow Hello").get().click();
+
+        waitForElement("Should have navigated to HelloWorld Flow",
+                By.id("flow-hello"));
+    }
+
+    @Test
+    public void hillaViewWithFlowOptOut() {
```
manolo, line 34: Both new tests are failing in CI, anything missing?

manolo (changes requested):

vaadin-platform-react-hybrid-test/src/test/java/com/vaadin/platform/react/test/FlowMainLayoutIT.java
ZheSun88: the failure is expected as it needs the fix (https://github.com/vaadin/hilla/pull/2934).

PR #7355, opened 2025-02-25
chore: disable feature flags in default bundles

mcollovati:
Updates vaadin-featureflags.js before running Vite to disable all feature flags in the bundle.

Releated to vaadin/flow#20991


mcollovati:
Closing, as we will follow another approach.

PR #7424, opened 2025-03-13
chore: fix error in workflow actions when assigning CE_LICENSE string in bash [skip ci]

manolo:
This fixes the '[: too many arguments' error when running pit preparation script

https://github.com/vaadin/platform/actions/runs/13815919957/job/38651445513#step:2:31

### DIFF
--- a/.github/workflows/pit.yml
+++ b/.github/workflows/pit.yml
@@ -63,7 +63,7 @@ jobs:
         run: |
           # COMPUTE MATRIX
           echo "Computing matrix ..."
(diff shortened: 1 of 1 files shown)


.github/workflows/pit.yml
```diff
@@ -88,6 +88,8 @@ jobs:
           M=`echo $M | sed -e s/,$//`']}'
           echo "Generated Matrix: $M"
           echo "matrix=$M" >> $GITHUB_OUTPUT
+        env:
+          CE_LICENSE: ${{secrets.CE_LICENSE}}
```
ZheSun88, line 92: do we still need the CE_license ? CE is a free product now.

manolo: Just for testing old branches, we can remove in a future

ZheSun88 (approved):

PR #7490, opened 2025-04-08
feat: add Master Detail Layout component to platform

sissbruecker:
## Description

Adds the Master Detail Layout component to the platform

## Type of change

- Feature

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -199,6 +199,12 @@
             "mode": "lit",
             "npmName": "@vaadin/login"
         },
+        "master-detail-layout": {
+            "javaVersion": "{{version}}",
+            "jsVersion": "24.8.0-alpha8",
+            "mode": "lit",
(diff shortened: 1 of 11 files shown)


versions.json
sissbruecker, line 1: Looks like the NPM package is not installed?
```
Failed to find the following imports in the `node_modules` tree:
- @vaadin/master-detail-layout/src/vaadin-master-detail-layout.js
If the build fails, check that npm packages are installed.
```

sissbruecker: It looks like it's caused by the exclusion in `versions.json` under `react-components`. Since the React component is not in alpha8, the component is thus not installed at all. So it seems https://github.com/vaadin/platform/pull/7495 needs to be merged first, which would install the component via react-components / bundles.

manolo (approved):

PR #7626, opened 2025-05-21
chore: upgrade WC to 25.0.0-alpha1

ZheSun88:


web-padawan:
Looks like `MPL-2.0` license is missing. It is used by [`dompurify`](https://github.com/cure53/DOMPurify) which is a `vaadin-markdown` dependency.

PR #7839, opened 2025-07-11
chore: pin jakarta version to 11.0.0 in vaadin-bom (25)

ZheSun88:


mcollovati (approved):

mcollovati:
Looking at #7770 it seems the main reason to use it was to have version managed in test modules.
So I think we can remove the JakartaEE BOM from the template and moving it into some other platform artifact

mcollovati:
#7841

PR #7866, opened 2025-07-24
chore: bump LC to 1.13.5 & update whitelists

manolo:


manolo:
close because already included in https://github.com/vaadin/platform/pull/7865

PR #8072, opened 2025-09-24
chore: remove Lumo iconset from bundles

sissbruecker:
Follow-up from https://github.com/vaadin/platform/pull/8068.

Removes the Lumo iconset from the default bundles, as those should not contain theme specific code. The iconset will be loaded automatically when using the `LumoIcon` class, however it will require building a custom bundle.


mshabarov (approved):

ZheSun88:
@mshabarov if the flow snapshot is ready, we can also update use flow snapshot version the `versions.json`. 
but i am afraid that the build will fail with the jackson incompatibility issue.

sissbruecker:
Closing this for now, this should be properly looked at and fixed by https://github.com/vaadin/flow/issues/23563

PR #8232, opened 2025-11-11
chore: upgrade testbench to 10.0.0-beta4

ZheSun88:


.github/workflows/sbom.yml
```diff
@@ -52,6 +52,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: '24'
+      - run: npm install -g npm@11.1
```
mshabarov: Is this really needed here?

ZheSun88: this is only for the sbom checking
and with the npm 11.2+, there is this issue. https://github.com/npm/cli/issues/8688

PR #8295, opened 2025-12-02
chore: upgrade org.owasp:dependency-check-maven version [skip ci]

ZheSun88:


ZheSun88:
sbom run https://github.com/vaadin/platform/actions/runs/19863273253

PR #8795, opened 2026-04-21
chore: upgrade CC and SSO-kit (24.10) [skip ci]

ZheSun88:


ZheSun88:
the goal of updating these two kits has met here.
https://github.com/vaadin/platform/actions/runs/24725562191

PR #8925, opened 2026-05-22
chore: remove badge and master-detail-layout feature flags

web-padawan:
## Description

These components will become stable in 25.2 so let's remove feature flags (WC and FC already updated).
Slider will be also removed in WC beta1 but it also needs Flow component PR, so I'll update it separately.

## Type of change

- Internal change

### DIFF
--- a/scripts/generator/templates/template-dev-bundle-pom.xml
+++ b/scripts/generator/templates/template-dev-bundle-pom.xml
@@ -54,14 +54,16 @@
         <!--
             vaadin-hybrid-dev-bundle must be a project dependency otherwise it will be built
             after vaadin-dev-bundle because of modules configuration in root pom.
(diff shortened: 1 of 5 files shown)


ZheSun88 (approved):

ZheSun88:
seems we also need to remove this from flow 
https://github.com/vaadin/flow/blob/main/flow-server/src/main/java/com/vaadin/experimental/FeatureFlags.java#L62
and https://github.com/vaadin/flow/blob/main/flow-server/src/main/java/com/vaadin/experimental/FlowComponentsFeatureFlagProvider.java

web-padawan:
@ZheSun88 created https://github.com/vaadin/flow/pull/24427 to remove MasterDetailLayout (Badge isn't listed there).

PR #8948, opened 2026-05-28
chore: list changed module in maintenance release note separately [skip ci]

### LINKED ISSUE #8938: Reduce Changelog for Unchanged Dependencies
### Describe your motivation

The current changelog makes it really hard to identify what has really changed between version because every component (even with no change) is added.

### Describe the solution you'd like

Only components (e.g. Flow / TestBench.. Kits..) are added to the change log if their version has changed. It would be cool to also show all intermediate releases in between - e.g.

Vaadin 25.1.6
- Flow (25.1.5 (old) -> 25.1.6 (intermediate) -> 25.1.7 (new))

ZheSun88:
fixes #8938 
currently, let us apply this change in the maintenance release only. 

for example, the main content for Vaadin 25.1.6 release note will be like (all version numbers will be linked to its release note)
image

### DIFF
--- a/scripts/generator/src/creator.js
+++ b/scripts/generator/src/creator.js
@@ -132,18 +132,121 @@ function createReleaseNotes(versions, releaseNoteTemplate) {
         }
     }
 
-    //const changed = getChangedSincePrevious(versions);
-    //console.log(versions.platform);
+    const changelogs = generateMaintenanceChangelog(versions);
     let releaseNoteData;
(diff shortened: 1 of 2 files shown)


knoobie:
Just an idea (if possible):

Add "Unchanged Modules" below with the current version so that people can easily find all modules?

--- 

Already thanks a lot! This looks so much better!

ZheSun88:
updated version 

image

PR #9077, opened 2026-06-26
Add vaadin-ee module

manolo:
Adds a new `com.vaadin:vaadin-ee` umbrella artifact, ee counterpart of `com.vaadin:vaadin`. It pulls the whole platform (through `vaadin-internal` and `vaadin-core-internal`) and additionally enforces a vaadin-ee license.


vaadin-ee/src/main/java/com/vaadin/platform/ee/EnterpriseEditionInitializer.java
```diff
@@ -0,0 +1,121 @@
...
+
+                ============================================================
+                   Running Vaadin Enterprise Edition {}
+                ============================================================
+                """, version);
+    }
+
+    private void logNoLicenseBanner(String productVersion) {
+        String version = orUnknown(productVersion);
+        LOGGER.warn("""
```
manolo: Not sure if warning here or error, wdyt @Legioth

vaadin-ee/pom.xml
```diff
@@ -0,0 +1,257 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ZheSun88: this file is not needed.. as will be generated ..

add it to .gitignore should be enough

pom.xml
ZheSun88, line 57: note for the release script/build update here:

need to add a flag to switch between `ee` and `oss` release. 
and no matter which release comes first, it should release the common modules (basically, the profile `platform` contains), and the other one should skip them.

ZheSun88: oh.. basically things mentioned in the BUILD.md.. good

vaadin-ee/pom.xml
manolo: done

ZheSun88:
i have modified the PR validation build to include the `oss` and `ee` profile.

scripts/generator/templates/template-vaadin-ee-pom.xml
```diff
@@ -0,0 +1,257 @@
...
+                        <goals>
+                            <goal>jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <includeDependencySources>false</includeDependencySources>
+                    <includeTransitiveDependencySources>false</includeTransitiveDependencySources>
+                    <quiet>true</quiet>
+                    <additionalparam>-Xdoclint:none</additionalparam>
```
mcollovati: This doesn't look like a valid parameter for `javadoc:jar` goal.

mcollovati (commented):
There are errors and warnings for the vaadin-ee module when running Javadoc plugin

```
[WARNING]  Parameter 'includeTransitiveDependencySources' is deprecated: if these sources depend on transitive dependencies, those dependencies should be added to the pom as
            direct dependencies
[INFO] The goal 'org.apache.maven.plugins:maven-javadoc-plugin:3.3.1:javadoc' has not been previously called for the module: 'com.vaadin:vaadin-core-internal:jar:25.2-SNAPSHOT'. Trying to invoke it...
[ERROR] MavenInvocationException: Error when invoking Maven, consult the invoker log file: /home/user/projects/vaadin/platform/platform/vaadin-ee/target/invoker/maven-javadoc-plugin1787258761.txt
[WARNING] Creating fake javadoc directory to prevent repeated invocations: /home/user/projects/vaadin/platform/platform/vaadin-core-internal/target/apidocs
[INFO] The goal 'org.apache.maven.plugins:maven-javadoc-plugin:3.3.1:javadoc' has not been previously called for the module: 'com.vaadin:vaadin-internal:jar:25.2-SNAPSHOT'. Trying to invoke it...
[ERROR] MavenInvocationException: Error when invoking Maven, consult the invoker log file: /home/user/projects/vaadin/platform/platform/vaadin-ee/target/invoker/maven-javadoc-plugin2019141350.txt
[WARNING] Creating fake javadoc directory to prevent repeated invocations: /home/user/projects/vaadin/platform/platform/vaadin-internal/target/apidocs
[ERROR] Error fetching link: /home/user/projects/vaadin/platform/platform/vaadin-core-internal/target/apidocs. Ignored it.
[ERROR] Error fetching link: /home/user/projects/vaadin/platform/platform/vaadin-internal/target/apidocs. Ignored it.
```

Not a blocker, but it would be good to fix.

mcollovati:
`vaadin-ee` is not in the Vaadin BOM. Is this on purpose?

vaadin-ee/src/main/java/com/vaadin/platform/ee/EnterpriseEditionInitializer.java
```diff
@@ -0,0 +1,121 @@
...
+        // no valid license, the normal banner otherwise.
+        if (isLicensed(version, productionMode)) {
+            logEnterpriseBanner(version);
+        } else {
+            logNoLicenseBanner(version);
+        }
+
+        // Delegate the real license handling (in dev mode this offers the trial
+        // through Dev Tools / Copilot; in production it does nothing).
+        super.serviceInit(event);
```
mcollovati, line 59: If a pre-trial is expired, the application does not start in dev mode.
An exception is thrown and the server stops, so it is not possible to see the splash screen and extend the trial by signing in.

mcollovati (changes requested):

vaadin-ee/src/main/java/com/vaadin/platform/ee/EnterpriseEditionInitializer.java
manolo: Thanks for catching this. I dug into the root cause:

The crash comes from the base class `BaseLicenseCheckerServiceInitListener` in flow-server, not from `EnterpriseEditionInitializer` itself. In dev mode with dev tools enabled it does:

```
try {
    LicenseChecker.checkLicense(productName, productVersion, buildType,
            url -> { throw new MissingLicenseKeyException(...); },
            0, Capabilities.of(Capability.PRE_TRIAL));
} catch (MissingLicenseKeyException ex) {
    // delegate to dev tools so the splash screen can be shown
    CheckProductLicense.register(event, productName, productVersion);
}
```

It only catches MissingLicenseKeyException. When there IS a local key but the pre-trial is expired, the checker throws PreTrialLicenseValidationException (a `LicenseValidationException`), which is not caught, so it propagates and aborts startup before `CheckProductLicense.register(...)` runs. That is why the app does not boot and the splash never appears to extend the trial.

The mechanism that shows the splash (`CheckProductLicense.register`) is private to the base class, so a subclass like EnterpriseEditionInitializer cannot reuse it without duplicating internal flow-server code, which would be fragile.

So I think the proper fix belongs in flow-server: in dev mode with dev tools enabled, catch `LicenseValidationException` (or the broader license-failure case) in addition to `MissingLicenseKeyException` and delegate to dev tools the same way, so an expired pre-trial still shows the splash instead of failing startup.

Want me to open a PR against flow for this? I can keep this platform PR as is and link them.

mcollovati: Yes, please go ahead with the Flow fix. I would catch for `PreTrialLicenseValidationException`.
Not sure if we also need to check that `getPreTrial().trialState` is `EXPIRED`, and only in that case and rethrow the exception for any other value.

manolo: done

scripts/generator/templates/template-vaadin-ee-pom.xml
manolo: done

mcollovati:
Actually, it looks a bit weird that `vaadin` and `vaadin-ee` use `vaadin-bom` in dependency management, and `vaadin-bom` pins `vaadin` and `vaadin-ee` versions

mcollovati (approved):

PR #9089, opened 2026-07-01
Trigger platform validation build (main)

manolo:
## Summary

Empty change to run the platform validation build against main and check whether the prod bundle comparison test behaves the same as on 25.2.


manolo:
Throwaway PR used to confirm the prod-bundle test behaves correctly on main. No longer needed.

PR #9125, opened 2026-07-09
chore: pin vaadin-themable-mixin in versions.json

web-padawan:
## Description

Same as https://github.com/vaadin/platform/pull/9109 but for `main` branch.

Versions are pinned during alpha / beta stage, but once we enter stable this can be a problem again.

## Type of change

- Internal change


manolo (approved):
Validation passed. Approving.

PR #9127, opened 2026-07-09
Stabilize ComponentsIT against slow app load

manolo:
## Summary

`ComponentsIT.appWorks` fails intermittently (on different browsers each run, across unrelated PRs) with a Selenium timeout at the first wait (`ComponentsIT.java:69`), while the error screenshots show the app still on the Vaadin loading indicator with blank content. The app just did not finish loading within the default 10s TestBench timeout.

Two changes:
1. Wait up to 30s for the initial notification instead of the default 10s, so a slow app load under CI load does not cause a spurious timeout.
2. Add `rerunFailingTestsCount` to the failsafe configuration so a flaky browser test is retried before failing the build.

## Context

The failure is not tied to any specific component or code change: it is the initial app load timing out under CI load. These changes make the browser test tolerant of that.

### DIFF
--- a/vaadin-platform-test/pom.xml
+++ b/vaadin-platform-test/pom.xml
@@ -240,6 +240,8 @@
                         </executions>
                         <configuration>
                             <forkCount>2</forkCount>
+                            <!-- Retry flaky browser tests (e.g. slow app load) before failing the build -->
+                            <rerunFailingTestsCount>2</rerunFailingTestsCount>
                         </configuration>
                     </plugin>
                     <plugin>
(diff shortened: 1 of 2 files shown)


manolo:
Closing: this does not fix the failure. ComponentsIT.appWorks fails because the app does not finish loading on Firefox/Safari via the internal hub in dev mode (parallel browsers), not because of a slightly short wait. The 30s wait still times out, and the retry is redundant since CI already passes -Dfailsafe.rerunFailingTestsCount=3 and TestBench maxAttempts=3. It is a platform wide infra/timing flake unrelated to any code change and needs a hub/dev-server side fix.

PR #9130, opened 2026-07-09
Allow selecting the tested browsers with -Dgrid.browsers on the hub

manolo:
## Summary

The `grid.browsers` system property was only read when running `ComponentsIT` against Sauce Labs. This makes it also apply when running against a Selenium hub (`com.vaadin.testbench.Parameters.hubHostname` set), so the browser set can be chosen from CI.

The default set (`firefox,firefox-128,safari-17,edge`) is unchanged when the property is not provided, so current behavior is preserved.

## Context

TeamCity PR validation runs these tests on the internal hub, where the property was previously ignored. With this change TeamCity can pass, for example, `-Dgrid.browsers=firefox-128,edge` to exclude browsers that are currently flaky on the grid, without touching the code or the Sauce runs.


manolo:
Not needed: ComponentsIT runs on Sauce Labs in PR validation, where grid.browsers is already honored. The tested browser set is now controlled from TeamCity via -Dsa.browsers.

PR #9142, opened 2026-07-11
chore: align kotlin and dokka versions with flow (#9133) (CP: 25.1)

mcollovati:
Flow 25.1 ships flow-gradle-plugin compiled with Kotlin 2.4.0 since vaadin/flow#24905, which the previous Kotlin 2.1.21 compiler cannot read (metadata 2.4.0). Picks only the build.gradle part of #9133, the versions.json hunk does not apply to this branch.


ZheSun88 (approved):
okay.. all good after getting a new flow 25.1 release

PR #9153, opened 2026-07-13
test: add Spring Boot reload-time benchmark

Artur-:
Adds vaadin-platform-reload-benchmark, a port of Flow's test-spring-boot-multimodule-reload-time benchmark. It measures Spring DevTools reload time of a large, generated Vaadin application. It lives here rather than in Flow because the point of the benchmark is to reload against the full platform component set, which only the platform repo can depend on.

The module is a four-part reactor mirroring the Flow original:
- generator: a Maven plugin that fabricates 500 routes plus Spring services, and (new here) an includeComponents flag that generates a route instantiating real platform components (Button, TextField, ComboBox, DatePicker, Grid, Checkbox) so the benchmark exercises component class-path scanning and frontend bundling. This replaces the Vaadin-24 third-party add-ons the Flow version used, which are not 25.x-compatible; depending on com.vaadin:vaadin already puts the entire first-party component set on the class path.
- library and theme: ship frontend resources and a theme from dependency jars, so the reloaded app resolves resources across modules.
- ui: the Spring Boot app plus the reload-time IT, depending on com.vaadin:vaadin. The IT runs in dev mode (DevTools) and is driven through a self-contained TestBench base class (AbstractReloadBenchmarkTest) instead of Flow's flow-test-util ChromeBrowserTest.

The module is gated behind a reload-benchmark profile in the root pom, so it runs the same way as the other platform IT modules (activated by PiT) rather than in the default build.


vaadin-platform-reload-benchmark/ui/pom.xml
```diff
@@ -0,0 +1,182 @@
...
+            <plugin>
+                <groupId>com.vaadin</groupId>
+                <artifactId>vaadin-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>prepare-frontend</goal>
+                        </goals>
+                    </execution>
+                </executions>
```
mcollovati: prepare-frontend is not needed

vaadin-platform-reload-benchmark/ui/pom.xml
```diff
@@ -0,0 +1,182 @@
...
+        <!-- The full platform component set: the scanning surface the
+             benchmark exercises. -->
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-spring</artifactId>
+        </dependency>
```
mcollovati, line 29: vaadin-dev dependency is missgin

```suggestion
        <dependency>
            <groupId>com.vaadin</groupId>
            <artifactId>vaadin-spring</artifactId>
        </dependency>
        <dependency>
            <groupId>com.vaadin</groupId>
            <artifactId>vaadin-dev</artifactId>
            <optional>true</optional>
        </dependency>
```

mcollovati (changes requested):

vaadin-platform-reload-benchmark/ui/pom.xml
mcollovati:
```suggestion
```

mcollovati:
Added a job on CI and it's now green. Comparing with the Flow repo module, the restart threshold raised from 1200ms to 1400ms because the platform module contains more dependencies (com.vaadin:vaadin instead or vaadin-core, vaadin-dev instead of vaadin-dev-server only, ...)

mcollovati (approved):

mcollovati:
And I guess the build failure is unrelated to this changes.

mcollovati:
By introducing this new module, the statistics charts on CI have to be split, since the tool does not support continuous lines for data points collected from different jobs.

Once this PR gets merged, we can go ahead with https://github.com/vaadin/flow/pull/24977

ZheSun88:
>And I guess the build failure is unrelated to this changes.

we need the new WC got merged in flow-component to process with platform PR validation.

PR #9157, opened 2026-07-14
chore: add Switch Component

ZheSun88:


vaadin-platform-react-hybrid-test/src/main/frontend/views/hilla/components.tsx
```diff
@@ -166,6 +167,7 @@ export default function Components() {
                 <Checkbox value="2" label="Option two"></Checkbox>
                 <Checkbox value="3" label="Option three"></Checkbox>
             </CheckboxGroup>
+            <vaadin-switch label="Notifications"></vaadin-switch>
```
web-padawan, line 170: This should use `Switch` when the next alpha with React component is published, let's wait for it.

vaadin-platform-react-hybrid-test/src/main/frontend/views/hilla/components.tsx
```diff
@@ -54,6 +54,7 @@ import {
     SideNavItem,
     Slider,
     SplitLayout,
+    Switch;
```
web-padawan: typo: should be `,` not `;`.

ZheSun88: done

vaadin-platform-react-hybrid-test/src/main/frontend/views/hilla/components.tsx
ZheSun88: okay.. 
current platform snapshot is red.. 
so we can make another PR to update it after the next WC release

ZheSun88:
aww
```
Error: [vite]: Rolldown failed to resolve import "@vaadin/switch/src/vaadin-switch.js" from "/home/runner/work/platform/platform/vaadin-prod-bundle/src/main/frontend/generated/flow/generated-flow-imports.js".

```

PR #9231, opened 2026-07-29
feat: report vaadin-ee usage in development mode

manolo:
## Summary

Register `vaadin-ee` with the current platform version through `UsageStatistics.markAsUsed` when `EnterpriseEditionInitializer` runs, so Enterprise Edition adoption shows up in the usage statistics.

The entry is registered in development mode only, since usage statistics are gathered exclusively in development (the client gatherer is stripped from production bundles and production instances are not tracked in the collector DB).

## Context

The identifier `vaadin-ee` plus the platform version rides Flow's existing usage statistics pipeline, so it lands in the collector as a normal element entry with no changes needed on the ingestion side.


ZheSun88 (approved):

ZheSun88:
This will be included in the next 25.3 alpha release (should be released today)

V25.2 will be shipped in the next v25.2 maintenance release (around 2 weeks)

PR #9235, opened 2026-07-31
test: add per-component theme CSS tests against real components

Artur-:
Flow's theme/components/<tag>.css feature injects styles into a component's shadow DOM via registerStyles from @vaadin/vaadin-themable-mixin, which only works on real ThemableMixin components. Testing it therefore needs a real Vaadin component, so these tests live in the platform repo rather than in Flow (where a plain faux element does not receive the registered styles).

Two modules under a new theme-component-css profile, both using real components from com.vaadin:vaadin:

- legacy-css-prod: a production build where three themes in a parent chain each ship components/vaadin-horizontal-layout.css with ::slotted(:nth-child(N)) rules; the IT asserts the three slotted children of a real vaadin-horizontal-layout render green/white/red.
- live-reload: a dev-mode app that adds/updates/deletes components/vaadin-text-field.css at runtime and asserts Vite HMR pushes the new border-radius into the real vaadin-text-field's shadow DOM on the fly.

The ITs run on a self-contained TestBench base (no flow-test-util dependency, to avoid pinning a flow test-artifact version).


ZheSun88:
enabled the test in the platform PR validaiton. seems there is [mismatched jackson version here](https://teamcity.vaadin.com/buildConfiguration/VaadinPlatform_VaadinPlatformPrValidation/352297?buildTab=log&focusLine=16256&logView=linear&linesState=3211.8833) 
```
com.fasterxml.jackson.core:jackson-annotations:jar: project version [2.18.3], plugin version [2.21]
[15:10:50 ](https://teamcity.vaadin.com/buildConfiguration/VaadinPlatform_VaadinPlatformPrValidation/352297?buildTab=log&focusLine=16257&logView=linear&linesState=16257)  [ERROR] The build process encountered an error: com/fasterxml/jackson/annotation/JsonSerializeAs com.fasterxml.jackson.annotation.JsonSerializeAs
[15:10:50 ](https://teamcity.vaadin.com/buildConfiguration/VaadinPlatform_VaadinPlatformPrValidation/352297?buildTab=log&focusLine=16258&logView=linear&linesState=16258)  [ERROR] To diagnose the issue, please re-run Maven with the -X option to enable detailed debug logging and identify the root cause.
```

PR #9362, opened 2026-09-02
chore: Update flow-cdi to 16.1.2 and vaadin-quarkus to 3.2.2

mshabarov:
## Summary

- flow-cdi 16.1.1 to 16.1.2
- vaadin-quarkus 3.2.1 to 3.2.2

## Context

Both bumps travel together so they take a single validation cycle. Replaces #9363.


manolo (approved):
Both TeamCity validations pass on b5f7d99e, Maven and Gradle, so the two bumps are validated together rather than one after the other.

Verified that both target artifacts are published in Maven Central: com.vaadin:vaadin-cdi:16.1.2 and com.vaadin:vaadin-quarkus:3.2.2. Note the versions.json key flow-cdi maps to the vaadin-cdi artifact.

The SBOM job is red and is not a required check on main. It was already failing on both original commits, 8e26c3b1 for the cdi bump and b3941898 for the quarkus bump, so it is pre-existing on this branch and not introduced here.

PR #9363, opened 2026-09-02
chore: Update vaadin-quarkus javaVersion to 3.2.2

mshabarov:


manolo (approved):

manolo:
Folded into #9362, which now bumps vaadin-quarkus to 3.2.2 as well, so both bumps take a single validation cycle.

PR #9366, opened 2026-09-02
chore: Update browserless-test to 1.2.0-beta1 and observability-kit-starter to 5.0.0-beta1

manolo:
## Summary

- browserless-test 1.2.0-alpha2 to 1.2.0-beta1
- observability-kit-starter 5.0.0-alpha1 to 5.0.0-beta1

## Context

A beta cannot ship alpha pins. Both versions are published and verified in Maven Central, and both carry their tag and GitHub release. Neither component's release build opens this PR, so the two pins travel together to spend a single validation cycle.

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -41,7 +41,7 @@
             "npmName": "@vaadin/breadcrumbs"
         },
         "browserless-test": {
-            "javaVersion": "1.2.0-alpha2"
+            "javaVersion": "1.2.0-beta1"
         },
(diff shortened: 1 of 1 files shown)


manolo:
Folded into #9368, which now carries the browserless-test and observability-kit-starter pins as well, so the three beta pins take a single validation cycle. #9368 is bot authored, which also keeps it approvable without a second reviewer.

